PR #27, opened 2018-05-20
Fix URLs for Swift files that were pointing to the forked repo

annjose:
Some of the URLs in the design docs were pointing to the forked repo. Those URLs have been updated to point to the `tensorflow` branch on the main repo.


annjose:
The commit in this PR was associated with an incorrect email address and cannot be modified. So I will create a new PR with a new commit.

PR #28, opened 2018-05-20
Fix URLs of Swift files that were pointing to the forked repo

annjose:
### What does this PR do?
This PR updates some of the URLs in the design docs that were pointing to the forked repo. Now they point to the `tensorflow` branch on the main repo.


dan-zheng:
Thank you!

PR #44, opened 2018-06-14
Use bullet operator • for matrix multiplication.

rxwei:
To be merged when https://github.com/apple/swift/pull/17173 is in and a new toolchain is built.

*Note:* On macOS, this can be easily typed with ⌥8.


lattner:
LGTM!

PR #47, opened 2018-07-11
lldb-api-based swift jupyter kernel

marcrasi:
This is a Jupyter Kernel for Swift that is implemented using LLDB's Python APIs.

This implementation should be more robust than talking to the Swift REPL using stdin/stdout (e.g. iSwift). I'm already getting pretty good outputs and error messages!

screenshot from 2018-07-10 17-39-57

This should eventually go in its own repo, but google's process for making new repos seems pretty heavyweight, so I thought I'd put it in here for now while it's experimental.

cc @rjpower


jupyter/README.md
```diff
@@ -0,0 +1,22 @@
...
+This is a Jupyter Kernel for Swift that is implemented using LLDB's Python APIs.
+
+# Installation Instructions
+
+Create a virtualenv and install jupyter in it.
+```
+virtualenv venv
+. venv/bin/activate
+pip -V # Make sure this says python 2.7. LLDB doesn't support python 3.
+pip install jupyter
```
rxwei: Most machines should have `pip2` command that guarantees python 2, no?

marcrasi: Updated !

lattner:
Very nice!!

PR #49, opened 2018-07-17
Added a proposal for a high-level design for graph program extraction.

eaplatanios:
This is based on some thoughts I had on this topic and on an offline discussion I had yesterday with @mhong and @alextp. I hope it leads to some interesting discussion. :)


docs/HighLevelDesignForGraphProgramExtraction.md
```diff
@@ -0,0 +1,70 @@
...
+
+1. Have a single entrypoint for the Swift program and a single entrypoint for the TensorFlow program (i.e., use a single TF graph for the whole program, and a single session that starts executing parts of the graph at the same time that the Swift program starts executing things on the Swift-side). _This is because creating multiple sessions can be expensive and also because resource sharing across sessions may not be easy to control (I am not 100% sure about this second comment, but it makes sense to me intuitively)_.
+2. Allow for communication between the two programs, wherever necessary, and make that communication explicit to the user of the API, without it being cumbersome to them, or requiring boilerplate code.
+3. Avoid the need for aggressive inlining of all TensorFlow-related code on the Swift-side (which is what is currently happening in the Swift implementation).
```
alextp: currently will get quickly outdated

docs/HighLevelDesignForGraphProgramExtraction.md
```diff
@@ -0,0 +1,70 @@
...
+2. ___TF -> Swift:___ `Tensor` objects allow for various ways to obtain their content, or a reference to them. One proposal could be to have methods like `.scalar`, `.toArray`, `.iterator`, and `.handle`, that return Swift-typed values. Whenever these are used, data needs to be transferred from TF to Swift and so we would add a `ReceiveFromTF` op to the graph. This actually again consists of two ops: (i) an `Enqueue` op, and a (ii) `Dequeue` op to a size 1 queue. The `Enqueue` op will be added to the TF graph, enqueueing the corresponding TF tensor (or its handle, for example) to the queue. At the same time, a `session.run(...)` call will be added to the Swift side, dequeueing the TF content that is to be received from the graph. These method calls will be required to be made explicitly from the library user and will not be added automatically by the Swift compiler, in order to avoid confusion and potential inefficiencies.
```
alextp: Hopefully you don't mean that the user will have to type "session.run" or something of the sort.

I also think this is missing discussion about how to just copy a pointer

docs/HighLevelDesignForGraphProgramExtraction.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Tacking Point 3: Handling Functions that use TF
+
+Currently, Swift for TF handles all Swift functions that use TF by aggressively inlining them. This can cause lots of issues, including an exploding size of the compiled code, as well as limited supported for recursion (note that TF graph functions themselves support recursion and thus we should be able to support it on our side too). Here we suggest we use TF graph functions to handle Swift functions that use TF. We propose the following high-level design: all Swift functions that use TF will be compiled into two separate functions:
+
+1. ___Swift Function:___ The Swift code part of the compiled function that contains all computation that needs to happen on the Swift side. Note that in the case that this is a pure TF function (using only TF ops), then this function could be compiled to a no-op, or optimized away entirely.
+2. ___TF Graph Function:___ The TF part of the compiled function that is a TF `FunctionDef` protobuf message which contains the extracted TF graph for the compiled function.
```
alextp: You should mention here what do to about the function signatures, and which information needs to be made available to swift and to tensorflow.

I think we can compute the swift-side and tf-side function signatures and store them somewhere as part of the type of this composite function. This needs to be visible at compile time so the compiler can build calls to tf functions without seeing its definition.

At link time (or runtime initialization, whatever is more convenient), when all the definitions are available, the full graph can be stitched together.

docs/HighLevelDesignForGraphProgramExtraction.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Desired Features
+
+As described in the GPE whitepaper, GPE is effectively a [program slicing](https://en.wikipedia.org/wiki/Program_slicing) method. This can be thought of as follows: working with Swift and TensorFlow means that we work with two different languages that get executed as two separate programs that can communicate when needed. One program is the main Swift program that also serves as the entrypoint of execution. The other is the _implicitly constructed_ TensorFlow graph that gets executed by the TensorFlow runtime. Ideally we would like to:
+
+1. Have a single entrypoint for the Swift program and a single entrypoint for the TensorFlow program (i.e., use a single TF graph for the whole program, and a single session that starts executing parts of the graph at the same time that the Swift program starts executing things on the Swift-side). _This is because creating multiple sessions can be expensive and also because resource sharing across sessions may not be easy to control (I am not 100% sure about this second comment, but it makes sense to me intuitively)_.
```
mhong: Agree this would be a really clean design! My immediate next step is to generate a single graph for each Swift module (but still retain the start graph / stop graph calls for each partitionable function). That should help us march towards the direction described here.

mhong: please feel free to remove the second comment and remarks in parenthesis -- simplicity itself (e.g. ease of debugging) is a strong reason to move towards a single session. :)

docs/HighLevelDesignForGraphProgramExtraction.md
```diff
@@ -0,0 +1,70 @@
...
+
+For all top-level code we check where `Tensor`s are used and how and we built a TF graph, based on the methods outlined in the GPE whitepaper. `Tensor`s that are created on the Swift-side are handled by the TF graph in the way described in the section tackling point 2. After building the TF graph in this manner, we collect all sink nodes in the TF graph, whether those are nodes that send tensors back to the Swift program, or simply nodes that only have a side-effect on the TF side, and we create a new TF `group` node that we call the `sink` node. This means that once the `sink` node finishes executing in TF, all TF-related computation for the current Swift for TF program will have finished executing and so it can be thought of as the TF _endpoint_.
+
+We have ignored functions that use TF for now, but we will discuss those in the section devoted to point 3.
+
+### Single TF Session
+
+Given the single TF graph we have built for the current Swift for TF program, we can then go ahead and add the following in the beggining of the top-level Swift program code (on the Swift side):
```
mhong: beggining has a typo

docs/HighLevelDesignForGraphProgramExtraction.md
```diff
@@ -0,0 +1,70 @@
...
+
+1. Create a new TF session (with any desirable configuration).
+2. Add a call to `session.run(sink)`.
```
mhong: if some sinks are in a control flow (if, while) context, we may not be able to run them in session.run() directly -- but it seems we can make the wrapping if/while nodes as sinks if needed (if their output values have no users).

docs/HighLevelDesignForGraphProgramExtraction.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Tackling Point 2: Communication Between Swift and TF
+
+Communication between Swift and TF can be handled in a clean manner, by following the same principles.
+
+1. ___Swift -> TF:___ The TF program deals with `Tensor` objects. Therefore, whenever a `Tensor` object is constructed on the Swift side, and later used by TF ops that require tensors, a corresponding `SendToTF` op is added to the graph. This actually consists of two ops: (i) an `Enqueue` op, and a (ii) `Dequeue` op to a size 1 queue. The `Dequeue` op will be used as the `Tensor` object representation for TF ops consuming it later on, and a `session.run(...)` call will be added to the Swift side, enqueueing the constructed tensor that will be sent to TF.
```
mhong: note "size 1 queue" can be relaxed and tuned. a larger sized queue could help with throughput, but might hurt latency (say we and sending / receiving a tensor value within a loop, we can send the values across many iterations without them being consumed, in some cases -- it's a producer/consumer buffer)

docs/HighLevelDesignForGraphProgramExtraction.md
mhong: the design of Tensor pointer/handle can probably be decpupled from this doc. let's follow up on that separately.

docs/HighLevelDesignForGraphProgramExtraction.md
eaplatanios: Which part are you referring to? My comment about resource management?

docs/HighLevelDesignForGraphProgramExtraction.md
eaplatanios: @alextp The compiler would add the call to `session.run`, and all other session management code. We should however provide a way for the user to pass their own `SessionConfig` if they want.

@mhong I agree. This document outlines very roughly some of the things we discuss about send/receive ops, but we should probably make these Moree concrete in a separate document. @alextp Regarding how to pass a pointer, there are a few ways we could do that, but the main thing we'd need to be careful about is reference counting when we pass tensor pointers around.

docs/HighLevelDesignForGraphProgramExtraction.md
eaplatanios: Richard's comment further down is relevant for this. I'll respond there.

docs/HighLevelDesignForGraphProgramExtraction.md
```diff
@@ -0,0 +1,70 @@
...
+
+1. ___Swift Function:___ The Swift code part of the compiled function that contains all computation that needs to happen on the Swift side. Note that in the case that this is a pure TF function (using only TF ops), then this function could be compiled to a no-op, or optimized away entirely.
+2. ___TF Graph Function:___ The TF part of the compiled function that is a TF `FunctionDef` protobuf message which contains the extracted TF graph for the compiled function.
+
+The calling code will be invoking both functions in parallel, in its two generated programs -- the Swift program and the TF program, respectively. Note that for the TF program, the invocation will consist of simply adding the corresponding `FunctionDef` to the graph function library and creating an op representing that function call.
+
+Here we omit details related to handling mixed Swift and TF arguments and also related to propagation such function calls which means that one function invocation may require multiple `FunctionDef`s to be added to the graph, as these can be handled relatively easily, and would distract from the main proposed idea.
```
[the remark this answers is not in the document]
eaplatanios: That's pretty much what I had in mind. We can also be more conservative about the SIL function type and let it be the full original function type (note that the tensor arguments they could be used in `ReceiveFromTF` calls in the Swift-side code of the function). In this case, if the tensor arguments are not used in the Swift-side code of the function, they will just be ignored and, for now, only incur a small cost of passing them in the stack (if that's not optimized away further down in the compilation process). Although, as you and Alex mention we could also just keep track of what's used where and remove arguments from either signature, as necessary.

In either case, I believe it would be a good idea to standardize this procedure and document it clearly (maybe in this document).

docs/HighLevelDesignForGraphProgramExtraction.md
eaplatanios: That sounds like a good first step that I think can start simplifying things. :)

eaplatanios: I will do that. I was also not sure if it's a good argument because it depends on how TensorFlow manages multiple sessions, which I am not familiar with.

docs/HighLevelDesignForGraphProgramExtraction.md
eaplatanios: Thanks for catching that! I keep confusing those double letters. :P

docs/HighLevelDesignForGraphProgramExtraction.md
rxwei: Session config is very much like the existing target config. It can be passed either via a top-level method call, like `TensorFlow.enableGPU()`, or via a compiler flag, like `-tf-enable-tpu`.

docs/HighLevelDesignForGraphProgramExtraction.md
eaplatanios: If the TensorFlow `If` and `While` ops produce output tensors, then we can just keep track of where those are used. If they do not, and we rely on side-effects (I'm not sure TF supports that behavior though), then we can consider that `If`/`While` op as a sink node itself and group it along with the other sink nodes. `tf.group` takes a set of ops as input and acts as a no-op barriers that waits for all of them to finish executing, and so should be able to handle this situation.

docs/HighLevelDesignForGraphProgramExtraction.md
eaplatanios: That's a good point! I will remove the size 1 constraint. :)

eaplatanios:
Thanks a lot for the comments everyone! I am glad you were also thinking of moving to something similar. I'll update with your comments today and I guess, one good next step would be to also think more about what the two parallel function signatures should look like (i.e., Swift and TF) and how to compute them.

docs/HighLevelDesignForGraphProgramExtraction.md
eaplatanios: Oh I see. Session config has many more options though than currently supported in the existing target config though, right? I guess these could be supported by adding similar top-level method calls for them. We might also want to support directly loading a `SessionConfig` proto, as I can imagine people serializing configurations for each server, in a distributed setting.

docs/HighLevelDesignForGraphProgramExtraction.md
mhong: Agree. The functional If and while ops should support "no output tensors" by design, though they are probably not exercised a lot, so some bugs could exist.

mhong:
Great! Agree that function signature needs some more design. @lattner mentioned this could be related to Swift's ABI (when compiling bar(), we don't know how it'll be called, so it must have a signature based on some convention), a subject that I don't know a lot about.

docs/HighLevelDesignForGraphProgramExtraction.md
rxwei: We need to be careful exposing proto-importing APIs for safety reasons, as loading a config proto at runtime won't provide early diagnostics.

A good first step is to add a compiler flag `-tf-session-config=<proto_file_path>` to compile the config proto along with the Swift program.

lattner:
I haven't had a chance to read the proposal in detail, but I'd prefer we make a new docs/ideas/ or /docs/proposals/ directory for new directions.  Any concerns with that?

lattner:
BTW, I'm very much +1 on the ideas in this doc as mhong@ can attest to :-)

rxwei:
+1 to having a proposal directory.

rxwei:
Proposals directory created https://github.com/tensorflow/swift/tree/master/proposals

eaplatanios:
I totally agree. I was actually not sure where to put this. :) Should we also decide on a basic template for bookkeeping purposes? For example, title/abstract/author/date, or something along these lines?

eaplatanios:
@rxwei I'll move this in the new directory.

rxwei:
You could follow the (Swift Evolution template](https://github.com/apple/swift-evolution/blob/master/proposal-templates/0000-swift-template.md). Sections like "source compatibility" and "effect on API resilience" won't make sense, so you can just delete them.

rxwei:
And how about a more explanatory file name, something like `OneSessionPerModule.md`?

mhong:
FYI - https://bugs.swift.org/browse/SR-8312?focusedCommentId=38036&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-38036 has some discussion on feeding tensors as input args of a graph, vs using host->TF sends after the graph execution starts. (You'll need to register a Jira account to view that Jira bug).

mhong:
(not sure if this is a good forum for in-depth design discussion, but I'm posting some thoughts here to get it started. Feel free to move the discussion to community mailing list, etc.)

One challenge in deciding function signature is that in order to produce an optimal compilation where the function arguments placement minimize sends/recvs, we'll need to do some "inter-procedural data flow analysis", rather than using just the argument data types. For a tensor-typed argument, it's possible that the function consuming it has host-only logic (e.g. running atari simulator on a tensor fully in swift, in an RL setting). 

For example, say when compiling function `foo`, we generate a pair of functions denoted as `{host(foo), graph(foo)}`. Say `foo` calls a function `bar` with some local tensor variable `x`, and `bar(x)` returns some tensor `y`.

For ABI stability, the calling convention will need to decide where to place `x` in the compiled version of `bar`, without looking into the body of `bar`. Say the decision is to place `x` in `graph(bar)` because of its tensor type. In that case `host(bar)` takes no arguments. Similarly, let `graph(bar)` return `y`.

In the scenario where `foo` and `bar` are host-only, the runtime workflow can be:
1. `host(foo)` sends its local variable `x` to `graph(foo)`
2. `graph(foo)` calls `graph(bar)` with `x` (recall `graph(bar)` is a no-op)
3. `graph(bar)` sends `x` to `host(bar)`, which computes `y`, sends it to `graph(bar)`, which then returns to `graph(foo)`
4. `graph(foo)` sends `y` to `host(foo)`, which continues its computation.

Optimizing away the sends/recvs above would be important for performance, but doing that after the modular partitioning and lowering as proposed here can seem inconvenient, or impossible without doing "global" (inter-procedural) analysis that involve all four generated functions above?

More generally, say `bar` consumes 10 input args instead of a single `x`, and `graph(bar)` computes 1 of them, while `host(bar)` computes the other 9. Assume `foo` is still host-only -- then we'll want to optimize away the sends/recvs over those other 9 params. How to do that?

Perhaps a middle-ground can be: only do "modular lowering", if `bar()` can be compiled to a graph-only function. Otherwise, we continue to do "global" dataflow analysis for partitioning/pruning first. Specifically, we inline `bar` into `foo` is when we compile `foo`, so that it turns inter-procedural data flow analysis into intra-procedural.

eaplatanios:
@mhong I see the point you raise, but as long as you package both the TF graph functions (i.e., the `FunctionDef`s) and the SIL code in the compiled libraries, what stops you from performing these global optimizations on the "client" library if you want to?

I also think there may be two topics of discussion here: (i) how to package libraries that use TF in a modular fashion, such that only the necessary TF graph functions are included in the "client" compiled binaries (i.e., to not include all the `FunctionDef`s defined in the library), and (ii) how to optimize the "client" code during compilation. But we could decouple those, as long as the packaging does not throw away any information, right?

eaplatanios:
One other question also came to my mind today while working through a possible Scala implementation. How do you plan to deal with op attributes? Only constant expressions should be allowed I assume. Is that right? I can definitely imagine cases where people provide a Swift function argument as an attribute to an op (e.g., `squeeze_dims` in the `Squeeze` op).

A more general question though is why do we want to even support attributes? Many ops are often revised to convert attributes to inputs so they can be dynamically provided, which is often useful. I get part where they are supposed to have statically-known values and that may help with graph construction, but shouldn't the TF graph compiler take care of that ideally? @alextp @lattner do you have any thoughts/plans about this?

rxwei:
> One other question also came to my mind today while working through a possible Scala implementation. How do you plan to deal with op attributes? Only constant expressions should be allowed I assume. Is that right? I can definitely imagine cases where people provide a Swift function argument as an attribute to an op (e.g., squeeze_dims in the Squeeze op).

Yes, today we use constant expression evaluation techniques to make sure arguments passed into op attribute positions are constant. If they aren't constant, an error will be emitted.

> A more general question though is why do we want to even support attributes? Many ops are often revised to convert attributes to inputs so they can be dynamically provided, which is often useful. I get part where they are supposed to have statically-known values and that may help with graph construction, but shouldn't the TF graph compiler take care of that ideally? @alextp @lattner do you have any thoughts/plans about this?

We do need attributes today, as some attributes specify the output dtype or the dtype of reduction indices, and some ops on axis are taking the axis as a attribute. Like you said, in the future, some attr-taking ops can be rewritten to take those attributes as inputs. What's missing here is an infrastructure to do partial specialization and dynamic graph generation. This is a very large topic, and would be better discussed and assessed on a separate thread/doc.

mhong:
@eaplatanios, I agree decoupling would be nice. But decoupled compilation would not be so useful, if optimization requires that we "throw away" the modular compilation output. e.g. say we've compiled some function `bar` into `graph(bar)` and `host(bar)`. When compiling `foo` which calls `bar`, if we will first inline `bar` into `foo` in order to do optimized partitioning, we are not benefiting from having `graph(bar)` and `host(bar)` available then.

rxwei:
@eaplatanios, when review comments are addressed, let's merge this to the repo and move discussions to the mailing list. Thanks!

rxwei:
Thanks. Merging now.

rxwei:
> If more optimizations are desired, then we would probably have to give away some of that modularity, but that does not have to happen on the library-side

Agreed. This is the case today in the `TensorFlow` module as well as many parts of the Swift standard library. Graph program extraction has the following rule:

* If a function is marked `@inlinable`, it is not being GPE'd. Instead, this will get inlined into whoever calls this function and get GPE'd as part of the caller.

The reasons behind this are:
- Efficiency. Most Swift TensorFlow ops are wrappers of `Raw` ops, which are wrappers of `#tfop`'s. We don't want to generate graph functions for these wrappers.
- Generics. If a function is generic (say, over the scalar type), there's no way to GPE it into a TF graph today. So `@inlinable` is required on all generic functions. (Note that partitioning generic functions is a large separate topic, and should be discussed on the mailing list later down the road.)

All APIs in the `TensorFlow` module are marked `@inlinable` and the code gets inlined into user code. The same strategy will apply to third-party libraries and won't conflict with the one-graph-function-per-function design proposed here.

mhong:
Agree with Anthony and Richard above.

BTW, just sent out https://github.com/apple/swift/pull/18282 to implement the "one graph per module" model as discussed above.

PR #55, opened 2018-08-12
Implicit Copy Warning Improvements

marcrasi:
Here's a proposal on how to clean up implicit copy warnings, based on some discussions with @mhong.

### DIFF
--- a/proposals/ImplicitCopyWarnings.md
+++ b/proposals/ImplicitCopyWarnings.md
@@ -88,7 +88,8 @@ public func train(inputs: Tensor<Float>, outputs: Tensor<Float>, initialWeights:
 We should emit a warning in the above code so that the user is aware that their
 training loop is blocking on data transfer and CPU computation.
 
-The round-trip-rule achieves exactly what we want!
+The round-trip-rule achieves exactly what we want in these examples! So does the
+heuristic.
 
 ## The round-trip-rule does not catch all slow programs
 
--- a/proposals/ImplicitCopyWarnings.md
+++ b/proposals/ImplicitCopyWarnings.md
@@ -4,32 +4,23 @@
 
 ## Introduction
 
-The Swift for TensorFlow compiler splits imperative tensor code into multiple
-programs that run on separate devices and occasionally send each other data.
-Since the splitting and sending happen implicitly, the compiler can generate
-programs with performance bottlenecks that the user does not expect (e.g. a
-program that sends data from the GPU to the CPU and then halts all GPU
-computation until the CPU processes the data and sends it back to the GPU). To
-address this problem, Swift for TensorFlow introduces implicit copy warnings
-that warn the user when data gets copied between devices. See the "Performance
-Predictability" section of [Graph Program Extraction] for more information.
-
-Currently, these warnings are very noisy. For example, [this simple model]
-produces [these warnings].
+Currently, implicit copy warnings are very noisy. For example, [this simple
+model] produces [these warnings]. (See the "Performance Predictability" section
+of [Graph Program Extraction] for more information about implicit copy
+warnings).
 
 I propose that we clean up the warnings as follows:
 1. Emit no warnings for data transferred while the program is starting (e.g.
    training data being copied to the GPU) or ending (e.g. final weights being
-   copied to the CPU). The current implementation might already intend to not
-   warn for these, but [the example] above demonstrates a lot of warnings at the
-   beginnings of partitioned functions.
+   copied to the CPU). [The example] demonstrates a lot of warnings at the
+   beginnings and ends of partitioned functions.
 2. Within the program, only warn when data makes a round trip from one device to
    another device and back again.
 
 Since the round-trip-rule might be hard to implement, I also propose that we
-initially implement a simple heuristic that approximates the round-trip-rule to
-try it out: Warn for all transfers from the host to the accelerator, but do not
-warn for any transfers from the accelerator to the host.
+initially implement a simple heuristic that approximates the round-trip-rule:
+Warn for all transfers from the host to the accelerator, but do not warn for any
+transfers from the accelerator to the host.
 
 Since all round trips involve a transfer from the host to the accelerator, the
 heuristic catches all transfers that the round-trip-rule catches.
@@ -72,12 +63,12 @@ public func train(inputs: Tensor<Float>, outputs: Tensor<Float>, initialWeights:
 }
 ```
 
-What we want to avoid is unexpectedly blocking users' computations on data
-transfers between devices. This can happen when the user writes some code that
-(unbeknownst to them) forces some computation to happen on the CPU before
-computation on the accelerator can proceed. For example, suppose that
-`cpuOnlyComputation` runs a computation that can only happen on the CPU. Then
-this training loop blocks on data transfer and CPU computation every iteration:
+What we want to avoid is unexpectedly blocking users' computations. This can
+happen when the user writes some code that (unbeknownst to them) forces some
+computation to happen on the CPU before computation on the accelerator can
+proceed. For example, suppose that `cpuOnlyComputation` runs a computation that
+can only happen on the CPU. Then this training loop blocks on data transfer and
+CPU computation every iteration:
 
 ```swift
 public func train(inputs: Tensor<Float>, outputs: Tensor<Float>, initialWeights: Tensor<Float>) -> Tensor<Float> {
@@ -99,8 +90,7 @@ public func train(inputs: Tensor<Float>, outputs: Tensor<Float>, initialWeights:
 We should emit a warning in the above code so that the user is aware that their
 training loop is blocking on data transfer and CPU computation.
 
-The round-trip-rule gives us exactly what we want in the example. So does the
-proposed heuristic.
+The round-trip-rule achieves exactly what we want!
 
 ## The round-trip-rule does not catch all slow programs
 
@@ -108,9 +98,11 @@ This rule obviously does not catch all slow programs. For example, a program
 that frequently dumps large pieces of data from the GPU to the CPU might soak up
 GPU memory with data waiting to be copied, and this rule will not catch that.
 
+TOD: Justify!
+
 ## Issues with the heuristic
 
-<DESCRIBE THESE!!!>
+TODO: Fill this in!
 
 We [recently eliminated warnings for scalar copies], because they cause a lot of
 noise when used for control flow. For example, the following code emits warnings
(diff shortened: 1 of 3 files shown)


proposals/ImplicitCopyWarnings.md
```diff
@@ -0,0 +1,144 @@
...
+   when a piece of data moves from device A to device B, then a computation
+   using that data happens on device B, and then the result of that computation
+   moves back to device A.
+
+Concretely, this proposal eliminates all warnings in [the example].
+
+Since the round-trip-rule might be hard to implement, I also propose that we
+initially implement a simple heuristic that approximates the round-trip-rule:
+Warn for all transfers from the host to the accelerator, but do not warn for any
+transfers from the accelerator to the host.
```
mhong: arguably warning only on swift->tf transfers can stand on its own right, rather than serving as an approximation due to impl-level convenience. e.g. say swift is running along side TF, and will send various tensors from swift to TF when these tensors are produced -- even without round-trips, such sends can slow down TF, if swift is being slow in producing these tensors.

proposals/ImplicitCopyWarnings/LinearRegression-warnings.txt
```diff
@@ -0,0 +1,84 @@
+reduced.swift:14:23: warning: 'inputs' implicitly copied to the accelerator, use .toAccelerator() to make transfer explicit
+  public func predict(inputs: Tensor<Float>) -> Tensor<Float> {
```
mhong: hmm i don't understand why such warnings appear in the first place -- there should be no sends/recvs here.
maybe this is due to the use of `mutating`?

proposals/ImplicitCopyWarnings.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Introduction
+
+Currently, implicit copy warnings are very noisy. For example, [this simple
+model] produces [these warnings]. (See the "Performance Predictability" section
+of [Graph Program Extraction] for more information about implicit copy
+warnings).
+
+I propose that we clean up the warnings as follows:
+1. Emit no warnings for data transferred while the program is starting (e.g.
```
mhong: agree #1 is a good idea, but how do we implement it (how do we tell whether a swift->tf data transfer happens _before_ the TF graph runs)?

Currently our use of "input args" and "result tensors" of the TF program serve the purpose of differentiating themselves from the sends/recvs, and thus generate no warning.

If we convert the input args and result tensors into tensor sends/recvs, i'm not sure how to reliably tell them apart from those other sends/revs that are happening _while_ the tf program is running.

Also, if swift is being too slow sending the first tensor to tf, it'd still increase the end-to-end running time, and it might be useful to warn in that case (if possible).

proposals/ImplicitCopyWarnings.md
```diff
@@ -0,0 +1,144 @@
...
+```
+
+If `step < steps` and `step % 2 == 0` get evaluated on the host and the boolean
+results get copied over to the accelerator, then the heuristic will warn about
+implicit copies. But these implicit copies are harmless because the host can
+quickly run through a bunch of iterations and queue up a bunch of booleans for
+the accelerator to consume.
+
+Without implementing the true round-trip-rule, we can denoise the warnings in
+that situation by suppressing warnings for scalar transfers. But the
```
mhong: and we already implemented the disabling off warnings on scalar values -- should we still include that in our policy design here (instead of forcing us to choose between it and the new round-trip-rule)?

proposals/ImplicitCopyWarnings.md
```diff
@@ -0,0 +1,144 @@
...
+round-trip-rule suppresses those warnings in a cleaner and more reliable way, so
+I propose that we eventually do implement the round-trip-rule.
```
mhong: another possible future direction that you mentioned in our earlier discussion is: 
Move sends/recvs diagnostics to a separate tool, possibly using runtime info (e.g.  profile data).

Do you want to also briefly cover it in this doc, to establish the "space" of our design exploration?

proposals/ImplicitCopyWarnings/LinearRegression-warnings.txt
rxwei: This is because `predict(inputs:)` is being partitioned as a deabstraction scope. It's a public function.

mhong: Before we have a better inlining model, should we recommend that all public APIs be marked as "inline always"? Otherwise if user code calls a sequence of public APIs, that could generate a lot of sends/recvs.

proposals/ImplicitCopyWarnings.md
```diff
@@ -0,0 +1,144 @@
...
+model] produces [these warnings]. (See the "Performance Predictability" section
+of [Graph Program Extraction] for more information about implicit copy
+warnings).
+
+I propose that we clean up the warnings as follows:
+1. Emit no warnings for data transferred while the program is starting (e.g.
+   training data being copied to the GPU) or ending (e.g. final weights being
+   copied to the CPU).
+2. Within the program, only warn when data makes a round trip between devices:
+   when a piece of data moves from device A to device B, then a computation
```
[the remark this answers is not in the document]
marcrasi: Oh yes, I meant to just talk about host<->accelerator transfers. I will reword this to make it more clear.

proposals/ImplicitCopyWarnings.md
marcrasi: I see this as similar to the tf->swift transfer case -- there are a lot of situations where the user does expect the transfers to happen and where the transfers don't slow anything down. For example, scalar transfers for control flow. Or some kind of CPU loading & preprocessing step that queues up a bunch of training data for the accelerator.

Thinking through scalars is what made me come up with the "round trip" idea. We can suppress all scalar warnings, like we do now, but this creates a possibly-bad situation in combination with suppressing tf->swift transfers: what if your program does a really slow tf->swift transfer and then sends it back as a scalar? Something like this:
```
for step in 0...1000 {
  let gradients = ... // on accelerator
  weights += gradients // on accelerator
  if cpuOnlyFunc(weights) == 0 { // long round-trip transfer that blocks the computation
    weights += 1
  }
}
```

If we hide all warnings for scalar transfers and we hide all warnings for tf->swift transfers, then that compiles without warnings. A "round trip" rule would catch it though.

proposals/ImplicitCopyWarnings/LinearRegression-warnings.txt
marcrasi: I think this discussion is mostly orthogonal to the warnings proposal. Regardless of what causes such a function to get partitioned on its own, we need to make the warnings good when it does get partitioned.

proposals/ImplicitCopyWarnings.md
marcrasi: I wanted to capture the current use of "input args" and "result tensors" in this rule, because it seems like part of the big picture.

Also it seems like they currently might be bugged in some way because a lot of the warnings in the example look like they should be input args or result tensors. So a big part of cleaning up the warnings might be debugging those.

Are we planning to eventually remove input args and result tensors and replace them with send/recv? If we do that, could we keep the logic that currently creates input args and result tensors, and use it to decide whether to warn?

proposals/ImplicitCopyWarnings.md
marcrasi: Will add!

proposals/ImplicitCopyWarnings.md
marcrasi: I think that this proposal interacts pretty interestingly with disabled scalar warnings, because of the example in my comment above with the slow scalar transfer.

proposals/ImplicitCopyWarnings/LinearRegression-warnings.txt
rxwei: Yep, it is orthogonal! But at the same time I hope this is providing some context on how argument send/receive warnings are emitted in the first place.

One middle ground I can think of is to have the compiler still partition inlinable functions outside the TF module. This would resolve both Mingsheng’s and mine concerns.

proposals/ImplicitCopyWarnings.md
mhong: If a tf-swift transfer is really slow, it might stall TF compute (due to limited TF RAM buffering) already, without involving a round-trip. :)

Overall what compiler can warn based on static analysis has limited accuracy, so we might as well go with a fairly simple (explainable / predictible to end users) model. Round-trip involves somewhat more mental calculation than saying "we warn on all host->tf transfers, other than scalar ones, but that's _still_ just an approximation". :)

But my position is based on some speculation, I'll let you decide here.

proposals/ImplicitCopyWarnings/LinearRegression-warnings.txt
mhong: Richard's proposal SGTM. 

I agree with Mark that what's partitionable is itself orthogonal to the warnings proposal. One related aspect is: when we are partitioning a public API, but if it does not get called in user code, it'd be nice not to generate sends/recvs warnings for it?

proposals/ImplicitCopyWarnings.md
mhong: > Also it seems like they currently might be bugged in some way because a lot of the warnings in the example look like they should be input args or result tensors. So a big part of cleaning up the warnings might be debugging those.

SGTM.

> Are we planning to eventually remove input args and result tensors and replace them with send/recv? If we do that, could we keep the logic that currently creates input args and result tensors, and use it to decide whether to warn?

I think that direction is worth exploration, but it's too early to commit to that. If we do go there, I agree that should probably not affect our warnings policy.

mhong:
Sorry about the delayed reply. The doc is good to land from my perspective!

rxwei:
@marcrasi Feel free to land this as is and await further feedback!

eaplatanios:
Thanks for putting together this document! I understand how such warnings can be annoying but I am leaning more towards letting the user know when data is moved without them explicitly specifying it. I also see how currently Swift for TF is a bit loose about devices; there is a notion of a single accelerator and host, but in many applications we use multiple GPUs, for example. Allowing for customized device placement is in the plan I imagine, and so assuming that eventually we will be able to place ops on devices as we choose, I believe that those warnings will be very useful. Currently, I feel I often lose control of when TensorFlow (using the Python API) will move stuff to the CPU when allowing soft placement and also when using "fake" int32 kernels. Therefore, I believe that the warnings are very useful even though they may be a bit verbose. And there is also a way to prevent them which is good.

One option may be to add such heuristics to prevent some of the warnings, but be able to not use them through say a compiler flag that controls verbosity for device placement warnings.

marcrasi:
Things have changed a lot since I wrote this. I think it's mostly obsolete, so I'll close it. Thanks for the comments everyone!

PR #62, opened 2018-08-29
⛳️add help for xcode 10 error

johndpope:


johndpope:
I see that's listed here - https://github.com/tensorflow/swift/blob/master/Usage.md 
Would be best to collapse these instructions.

PR #92, opened 2019-01-04
🏗XcodeGen - automagically build new xcode projects using swift-latest symlink

johndpope:


johndpope:
the toolchain path can be easily switched for different toolchains. 
see this updated sample.
https://gist.github.com/johndpope/b91748d307b2a49e710d8d0580bd52a8

marcrasi:
Thanks for the contribution. Since we didn't look at this in January, it may be quite outdated by now. We'll close it for now.

PR #98, opened 2019-01-20
Add link for SSA

lamberta:
For static single assignment (SSA) since this wasn't the easiest term to google for in this context.


rxwei:
Thanks!

PR #105, opened 2019-02-05
Updating installation instructions to point to latest builds.

bgogul:
Don't have a way to update the date of the build yet. We should perhaps have our automated builds updated this fie directly.

This also removes references to Ubuntu 14.04 and 16.04 for now. Once we have an automated builds, we should put them back. 

Also, this does not update the installation instructions. I will do so in a subsequent PR.

### DIFF
--- a/Installation.md
+++ b/Installation.md
@@ -10,7 +10,7 @@ To install Swift for TensorFlow, download one of the packages below and follow t
 
 Packages will be released nightly after automated building is set up.
 
-| Download | Date [(last commit)](https://github.com/apple/swift/commit/280486afdc0c940af345c614a80826fcc8326abc) |
+| Download | Date |
 |----------|------|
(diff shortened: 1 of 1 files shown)


Installation.md
```diff
@@ -13,8 +13,8 @@ Packages will be released nightly after automated building is set up.
 | Download | Date [(last commit)](https://github.com/apple/swift/commit/280486afdc0c940af345c614a80826fcc8326abc) |
```
marcrasi: The "last commit" on this line is no longer accurate. Could you remove it?

marcrasi:
These changes LGTM. I think you should wait until you have verified and updated the installation instructions before merging this though.

bgogul:
@marcrasi I was able to install the cuda 10.0 tar on a fresh ubuntu 18.04 and test it. I have updated the instructions accordingly.

Installation.md
bgogul: Done,

PR #115, opened 2019-02-16
Create LayerProtocolRevision.md

ewconnell:
Proposed changes the the Layer.applied(to: function

### DIFF
--- a/proposals/LayerProtocolRevision.md
+++ b/proposals/LayerProtocolRevision.md
@@ -1,0 +1,172 @@
+# Layer Protocol - Training and Concurrent Testing
+
+* Author: [@ewconnell](https://github.com/ewconnell)
+
+## Introduction
+This document discusses the requirements and proposed design changes to enable concurrent testing while training. It also discusses problems with use of the _LearningPhaseIndicator_ class in the current design.
+
+## Performance
+Significant training time improvement can be achieved by performing model test passes concurrently, allowing training to continue uninterrupted. The larger the training sample, number of samples in the test set, and more expensive the model design, the greater the benefit. Concurrent test passes can utilize idle GPU capacity, utilize additional GPUs, or be distributed to other nodes.
+
+The current S4TF implementation does allow this to a limited degree. The following example runs correctly. In this case very little benefit is achieved because the S4TF training pass is currently much slower than the inference pass. In Netlib a training pass is very fast, so I was able to confirm that a concurrent test pass achieves a significant performance gain. The following is an example of a concurrent training loop that currently works with the simple addition of a _DispatchQueue_ and _DispatchGroup_. Since Layer is a struct, models copy correctly. As training proceeds the learned parameters mutate, making the copies independent.
+```swift
+var model = MNISTClassifier()
+let optimizer = SGD<MNISTClassifier, Float>(learningRate: 0.1, momentum: 0.9)
+let batchSize: Int32 = 60
+let testBatchSize: Int32 = 1000
+let trainingIterations: Int32 = trainingImages.shape[0] / batchSize
+let epochs = 10
+let testQueue = DispatchQueue(label: "testQueue")
+let testGroup = DispatchGroup()
+
+func minibatch<T>(_ x: Tensor<T>, size: Int32, batchIndex: Int32) -> Tensor<T> {
+  let start = batchIndex * size
+  return x[start..<start + size]
+}
+
+print("Begin training for \(epochs) epochs" )
+let start = Date()
+
+for epoch in 0..<epochs {
+  var totalLoss: Float = 0
+  // train
+  for i in 0..<trainingIterations {
+    let images = minibatch(trainingImages, size: batchSize, batchIndex: i)
+    let labels = minibatch(oneHotTrainingLabels, size: batchSize, batchIndex: i)
+
+    let gradients = gradient(at: model) { model -> Tensor<Float> in
+      let logits = model.applied(to: images)
+      let batchLoss = softmaxCrossEntropy(logits: logits, labels: labels)
+      totalLoss += batchLoss.scalarized()
+      return batchLoss
+    }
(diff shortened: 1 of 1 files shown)


ewconnell:
I signed it!

proposals/LayerProtocolRevision.md
```diff
@@ -0,0 +1,172 @@
...
+public protocol Layer: Differentiable & KeyPathIterable
+    where AllDifferentiableVariables: KeyPathIterable {
+    /// The input type of the layer.
+    associatedtype Input: Differentiable
+    /// The output type of the layer.
+    associatedtype Output: Differentiable
+
+    /// Returns the output obtained from applying to an input.
+    @differentiable(wrt: (self, input))
+    func applied(to input: Input, training: Bool) -> Output
```
saeta:
```suggestion
    func applied(to input: Input, in context: Context) -> Output
```

where `Context` is defined as (something similar to):

```swift
protocol Context {
  var training: Bool { get }
}
```

The reason being is that: I think it'd be a good idea to make this a bit more general / flexible. I suspect that it may be valuable for layers to (eventually) be able to know a bit more about the context within which they're executing. e.g. if we're fine-tuning for quantization. This way, if we decide we need to have an additional orthogonal value (or if we decide that there's actually more than 2 modes (e.g. not just training and inference but say fine-tuning) we have the flexibility to make that change without causing a lot of churn.

The risk I see with going in this direction, however, is that the `Context` object could easily become a dumping ground for arbitrary utility things. So, it'd be very good we are careful.

Additionally, I think there's an open question as to whether `context` should be the first parameter or the second parameter. (e.g. Golang as a similar pattern, and always requires `Context` to be the first value.)

Finally, I've suggested `Context` to be a protocol, but I'm not sure if it would better be served as a `struct`. I imagine this decision has some flexibility tradeoffs, as well as performance implications. I'd be curious to hear more from those who have carefully tuned performance in Swift.

What do folks think of this direction?

ewconnell: I think defining a Context protocol opens the door to forward flexibility and is a reasonable idea. If the notion of a Context is adopted then I agree stylistically it would be best as the first parameter.

Models are intended to run in a variety of execution contexts, so this could be an ideal way to bundle up and pass through a collection of control parameters from User level to the back end. For example there are a variety of parameters that can be used to specify Cuda behavior, and in the GCP case there might be a completely different set of context information for accounts, specific resources, and resource configuration parameters.

A set of objects conforming to _Context_ with additional specialized protocols for each of the common environments (CPU, Cuda, GCP) could be included as part of the product to cover 99% of the user scenarios. The collection could be added to over time if needed and passed through without requiring modifications to the rest of the API.

The compiler seems to do a pretty good job of efficiently passing structs down a call tree. I believe even though they are value types, the compiler simply passes a ref until there is an attempt to mutate, which I don't expect would happen other than at initialization time.

Your concern about the context being abused and becoming a dumping ground for poorly designed 3rd party components is quite valid. I don't know what the company policy is about asserting control over things. I expect most users will have no need to use anything other than the standard set. In any case, clear guidelines should be provided to discuss the issues involved in creating a custom context.

proposals/LayerProtocolRevision.md
```diff
@@ -0,0 +1,172 @@
...
+}
+
+public extension Layer {
+  func applied(to input: Input) -> Output {
+    return applied(to: input, training: false)
```
rxwei: I don't think it could lead to unexpected behavior make this method available. For instance, you could pass `batchNorm.applied(to: x)` without a second argument in a model's `applied(to:)` method, but it could be misleading and cause unintended behavior even though it looks okay. Moreover, semantically, sublayers' learning phase should be training, when the parent `applied(to:)` is training, so passing `false` would be semantically inconsistent.

saeta: +1 to making the context parameter required (and not defining this extension). It'd probably be a code smell to not pass through the context / generate a new context implicitly. Thanks!

rxwei:
I created https://github.com/tensorflow/swift-apis/pull/13 that implements this. However it's blocked by a compiler bug. We working on a fix.

ewconnell:
When I started this I was trying to suggest a simple change for a single parameter so that models could be copied correctly. After our discussion and your valuable input, I think the learning phase is only a small part of the larger issue of execution context and resource initialization. I have some experience with this problem, and thoughts on requirements. The Dev Summit is coming soon, and I don't believe this larger set of issues will get fully written up, reviewed, implemented, and tested by then.  @saeta Brennan, would you like me to take a first pass at writing up the requirements, and then proposing an implementation? No worries if this is something you guys think you would prefer to address yourselves.

Ed

rxwei:
I think cleaning up the learning phase issue is a very important first step and should land ASAP. I'm fixing the blocker that's preventing it. The current design along with models built with it is kind of an anti-pattern right now.

ewconnell:
One other note, I don't think Context can be a class as suggested. It would create the same model copying problem as LearningPhaseIndicator. The class pointer would be copied preventing concurrent training and testing. I believe a final solution will also involve being able to "set" (change) the context of a copied model. It would be ideal to have the flexibility to train on the GPU and concurrently test a model on the CPU using both compute resources. Similarly you would probably want to train on a GCP TPU and perhaps offload testing to another GCP CPU node, etc..

ewconnell:
With the emphasis on doing something in time, I am currently working on the modifications to Layer.swift for your review.

rxwei:
My intuition is to make any copying of `Context` be explicit, but I could be wrong.  When a layer passes a context to a sublayer, semantically it should be borrowed, not copied.

ewconnell:
I've checked the changes into my forked execution-context branch, however it doesn't fully build yet because valueWithPullback is being used by BatchNorm and Dropout, so I can't simply remove them as previously suggested. I can create a pull request so that you have access to my changes, however I didn't know if that is acceptable since it doesn't cleanly build yet?

PR #116, opened 2019-02-19
Add toolchain without GPU support to installation instructions.

bgogul:
Fixes: https://bugs.swift.org/browse/TF-206


rxwei:
How about just `(CPU)`?

saeta:
How about `(CPU only)`? :-)

bgogul:
Changed to CPU Only. :)

PR #137, opened 2019-03-08
Add differentiable types doc.

dan-zheng:
Add "Differentiable types" document, explaining the design of the
`Differentiable` protocol.

### DIFF
--- a/docs/DifferentiableTypes.md
+++ b/docs/DifferentiableTypes.md
@@ -134,7 +134,7 @@ Here is a detailed explanation of the `Differentiable` protocol:
 * `associatedtype CotangentVector` represents the type of gradient values computed via reverse-mode differentiation.
   * `CotangentVector` types are used and produced by differential operators like `gradient` and `pullback`.
 * `var allDifferentiableVariables: AllDifferentiableVariables` represents all differentiable variables in an instance of the conforming type, where `associatedtype AllDifferentiableVariables` is the type of all differentiable variables.
-  * The motivation/design behind "all differentiable variables" is enabling key-path-based parameter optimization by making parameters and their gradients have the same type. Read the [synthesis rules](#compiler-synthesized-implementations) below and the [parameter optimization doc][parameter-optimization] for more information.
+  * The motivation/design behind "all differentiable variables" is enabling key-path-based parameter optimization by making parameters and their gradients have the same type. Read the [synthesis rules](#compiler-synthesized-implementations) below and the [parameter optimization document][parameter-optimization] for more information.
 * `TangentVector`, `CotangentVector`, and `AllDifferentiableVariables` are closely related.
   * All three associated types must themselves conform to `Differentiable`.
   * The `Differentiable` protocol associated types of the associated types themselves are defined to be mathematically correct.
--- a/docs/DifferentiableTypes.md
+++ b/docs/DifferentiableTypes.md
@@ -1,0 +1,293 @@
+# Differentiable types
+
+[Richard Wei](https://github.com/rxwei), [Dan Zheng](https://github.com/dan-zheng)
+
+Last updated: March 2019
+
+## Preface
+
+Speaking in terms of elementary calculus, only functions are "differentiable": only functions have derivatives and can be differentiated. In this document, the terminology "differentiable types" is used as a shorthand for "types that can be used as arguments and results of differentiable functions". This notion is important because not all types are "differentiable" in this sense. For example, types representing real numbers and vector spaces are "differentiable", but strings and integers are not.
+
+## Introduction
+
+Elementary calculus defines differentiation on real numbers: most people are familiar with this definition of "differentiation". However, differentiation is defined for many concepts across different branches of mathematics:
+* Scalar differentiation: differentiation on real numbers. This is taught in introductory calculus.
+* [Vector calculus]: a branch of mathematics that involves differentiation of vector fields.
+* [Differential geometry]: a branch of mathematics that involves differentiation of functions over manifolds.
+
+In Swift, we want to build a general system for differentiation that can represent all of these cases. Differentiation should not be limited to functions over specific types (e.g. functions over floating-point numbers); it should also work with functions whose parameters/result are custom types.
+
+This raises some questions: what kind of types can be used as arguments and results of differentiable functions, and how can we generalize them using a protocol?
+
+## Design overview
+
+The `Differentiable` protocol generalizes all types that can be used as arguments and results of differentiable functions.
+
+The compiler can automatically provide default implementations of `Differentiable` protocol requirements for struct types whose stored properties all conform to `Differentiable`.
+
+Here are some examples:
+
+```swift
+struct Vector: Differentiable, VectorNumeric {
+    // The compiler synthesizes all `Differentiable` protocol requirements
+    // when all stored properties conform to `Differentiable`.
+    var x, y, z: Float
+}
+
+// Differential operators like `gradient(at:in:)` just work!
+let v = Vector(x: 1, y: 2, z: 3)
+let 𝛁v = gradient(at: v) { v in (v + v).x }
+
+print(𝛁v)
+// Vector(x: 2.0, y: 0.0, z: 0.0)
+```
+
+A `Differentiable`-conforming type may have stored properties that are not meant to have a derivative with respect to `self`. Use the `@noDerivative` attribute to mark those properties; they will not have a corresponding entry in the synthesized `TangentVector`, `CotangentVector`, and `AllDifferentiableVariables` struct types.
+
+Here’s an example deep learning layer with some `@noDerivative` properties:
+
+```swift
+struct DenseLayer: Differentiable {
+    // These properties should have derivative values.
+    var weight: Tensor<Float>
+    var bias: Tensor<Float>
+
+    // These auxiliary properties should not have derivative values.
+    // Thus, they are marked with `@noDerivative`.
+    //
+    // `@noDerivative` properties do not have a corresponding entry in synthesized associated struct
+    // types.
+    @noDerivative var useBias: Bool = true
+    @noDerivative var previousWeight: Tensor<Float> = Tensor(0)
+
+    init(weight: Tensor<Float>, bias: Tensor<Float>) {
+        self.weight = weight
+        self.bias = bias
+    }
+
+    // The compiler synthesizes all `Differentiable` protocol requirements, adding only properties
+    // not marked with `@noDerivative` to associated tangent space types.
+
+    func applied(to input: Tensor<Float>) -> Tensor<Float> {
+        return matmul(input, weight) + bias
+    }
+}
+
+// Differential operators like `gradient(at:in:)` just work!
+let dense = DenseLayer(weight: [[1, 1], [1, 1]], bias: [0, 0])
+let 𝛁dense = gradient(at: dense) { dense in dense.applied(to: [[3, 3]]).sum() }
+
+dump(𝛁dense)
+// ▿ DenseLayer.AllDifferentiableVariables
+//   - weight: [[3.0, 3.0], [3.0, 3.0]]
+//   - bias: [1.0, 1.0]
+```
+
+## Protocol details
+
+Here is the full `Differentiable` protocol definition. More explanation is provided below.
+
+```swift
+/// A type that mathematically represents a differentiable manifold whose
+/// tangent spaces are finite-dimensional.
+public protocol Differentiable {
+    /// The tangent bundle of this differentiable manifold.
+    associatedtype TangentVector: AdditiveArithmetic & Differentiable
+        where TangentVector.TangentVector == TangentVector,
+              TangentVector.CotangentVector == CotangentVector,
+              TangentVector.AllDifferentiableVariables == TangentVector
+    /// The cotangent bundle of this differentiable manifold.
+    associatedtype CotangentVector: AdditiveArithmetic & Differentiable
+        where CotangentVector.TangentVector == CotangentVector,
+              CotangentVector.CotangentVector == TangentVector,
+              CotangentVector.AllDifferentiableVariables == CotangentVector
+    /// The type of all differentiable variables in this type.
+    associatedtype AllDifferentiableVariables: Differentiable
+        where AllDifferentiableVariables.AllDifferentiableVariables == AllDifferentiableVariables,
+              AllDifferentiableVariables.TangentVector == TangentVector,
+              AllDifferentiableVariables.CotangentVector == CotangentVector
+
+    /// All differentiable variables in this type.
+    var allDifferentiableVariables: AllDifferentiableVariables { get }
+
+    /// Returns `self` moved along the value space towards the given tangent vector.
+    /// In Riemannian geometry (mathematics), this represents an exponential map.
+    func moved(along direction: TangentVector) -> Self
+
+    /// Convert a cotangent vector to its corresponding tangent vector.
+    func tangentVector(from cotangent: CotangentVector) -> TangentVector
+}
+```
+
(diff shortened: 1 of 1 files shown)


docs/DifferentiableTypes.md
```diff
@@ -0,0 +1,281 @@
...
+## Protocol details
+
+Here is the full `Differentiable` protocol definition. More explanation is provided below.
+
+```swift
+/// A type that mathematically represents a differentiable manifold whose
+/// tangent spaces are finite-dimensional.
+public protocol Differentiable {
+    /// The tangent bundle of this differentiable manifold.
+    associatedtype TangentVector : AdditiveArithmetic & Differentiable
```
rxwei:
```suggestion
    associatedtype TangentVector: AdditiveArithmetic & Differentiable
```
Same elsewhere.

docs/DifferentiableTypes.md
```diff
@@ -0,0 +1,289 @@
...
+  <br>
+  <sub>From https://en.wikipedia.org/wiki/Pushforward_(differential).</sub>
+</p>
+
+Here are the important points about the `Differentiable` protocol:
+* `associatedtype TangentVector`: represents the type of directional derivatives computed via forward-mode differentiation.
+* `associatedtype CotangentVector`: represents the type of gradient values computed via reverse-mode differentiation.
+* `CotangentVector` types are used and produced by differential operators like `gradient` and `pullback`.
+* `var allDifferentiableVariables: AllDifferentiableVariables`: represents all differentiable variables in an instance of the conforming type, where `AllDifferentiableVariables` is the type of all differentiable variables.
+  * The motivation/design behind this is related to key path iteration .
```
dan-zheng: Todo: complete this unfinished sentence.

docs/DifferentiableTypes.md
```diff
@@ -0,0 +1,289 @@
...
+    /// vector. In Riemannian geometry (mathematics), this represents an
+    /// exponential map.
+    func moved(along direction: TangentVector) -> Self
+
+    /// Convert a cotangent vector to its corresponding tangent vector.
+    func tangentVector(from cotangent: CotangentVector) -> TangentVector
+}
+```
+
+Mathematically, `Differentiable` represents a differentiable manifold: this is a technical term for smooth-surfaced objects like spheres and generalizes types that are compatible with differentiation, like `Float`, `Double`, `Tensor`, and `SIMD4<Float>`. This definition comes from differential geometry and is quite technical, and not all details are relevant for most use cases.
```
rxwei: This is the first mention of differentiable manifold. Add a link?

docs/DifferentiableTypes.md
```diff
@@ -0,0 +1,289 @@
...
+
+<p align="center">
+  <img src="https://upload.wikimedia.org/wikipedia/commons/3/37/Pushforward.svg" align=center>
+  <br>
+  <sub>Image showing two differentiable manifolds: a sphere and a spheroid.</sub>
+  <br>
+  <sub>From https://en.wikipedia.org/wiki/Pushforward_(differential).</sub>
+</p>
+
+Here are the important points about the `Differentiable` protocol:
```
rxwei: Hmm... "Points" sounds pretty weird to me since you are iterating over every requirement and explaining them.

docs/DifferentiableTypes.md
```diff
@@ -0,0 +1,289 @@
...
+
+```swift
+extension Float: Differentiable {
+    public typealias TangentVector = Float
+    public typealias CotangentVector = Float
+    public typealias AllDifferentiableVariables = Float
+}
+// The conformance for `Double` is defined similarly.
+
+extension Tensor: Differentiable where Scalar : TensorFlowFloatingPoint {
```
rxwei:
```suggestion
extension Tensor: Differentiable where Scalar: TensorFlowFloatingPoint {
```

docs/DifferentiableTypes.md
```diff
@@ -0,0 +1,289 @@
...
+`var allDifferentiableVariables: AllDifferentiableVariables` is synthesized as a computed property that mirrors the differentiation properties of the conforming type.
+
+It is always synthesized with a getter.
+It is synthesized with a setter only when all differentiation properties are mutable and themselves all have mutable `allDifferentiableVariables` properties.
+
+Examples:
+```swift
+// Example when `AllDifferentiableVariables == Self`.
+var allDifferentiableVariables: AllDifferentiableVariables {
+    get { return self }
```
rxwei: Speaking of this, we should totally change derived conformances to derive `_read` and `_modify` accessors.

docs/DifferentiableTypes.md
```diff
@@ -0,0 +1,289 @@
...
+  * All three associated types must themselves conform to `Differentiable`.
+  * The `Differentiable` protocol associated types of the associated types themselves are defined to be mathematically correct.
+    * `Foo.TangentVector.TangentVector` is `Foo.TangentVector` itself.
+    * `Foo.CotangentVector.TangentVector` is `Foo.CotangentVector` itself.
+    * `Foo.TangentVector.CotangentVector` is `Foo.CotangentVector`.
+    * `Foo.CotangentVector.CotangentVector` is `Foo.TangentVector`.
+    * `Foo.AllDifferentiableVariables` has the same `TangentVector` and `CotangentVector` as `Foo`.
+  * Additionally, `TangentVector` and  `CotangentVector` must conform to `AdditiveArithmetic`, so that they can be zero-initialized and accumulated via addition. These are necessary to perform the chain rule of differentiation.
+* Manifold operations.
+  * These currently involve `func tangentVector(from cotangent: CotangentVector)` and `func moved(along cotangent: CotangentVector`). These operations can be useful for implementing manifold-related algorithms, like optimization on manifolds, but are not relevant for simple differentiation use cases.
```
rxwei: * `moved(along:)` takes a tangent, not a cotangent.
* If you write the function decl, then it would be more appropriate to write the return type as well. Otherwise, using their short names seems better.

```suggestion
  * These currently involve `tangentVector(from:)` and `moved(along:)`. These operations can be useful for implementing manifold-related algorithms, like optimization on manifolds, but are not relevant for simple differentiation use cases.
```

docs/DifferentiableTypes.md
```diff
@@ -0,0 +1,289 @@
...
+
+The standard library defines conformances to the `Differentiable` protocol for `Float` and `Double`, as well as a conditional conformance for `Tensor`:
```
rxwei: The standard library doesn't have `Tensor`, but we should probably mention SIMD vector types.

```suggestion
The standard library defines conformances to the `Differentiable` protocol for `Float`, `Double`, and `Float80`, as well as conditional conformances for SIMD vector types.
```

docs/DifferentiableTypes.md
rxwei: Not sure why we should mention `Tensor` here. Can we define a new `Vector` type to demonstrate this?

docs/DifferentiableTypes.md
```diff
@@ -0,0 +1,289 @@
+# Differentiable types
+
+Richard Wei, Dan Zheng
```
rxwei: Add github links.

docs/DifferentiableTypes.md
```diff
@@ -0,0 +1,289 @@
...
+
+March 2019
+
+## Preface
+
+Mathematically speaking, only functions are "differentiable": only functions have derivatives and can be differentiated. In this document, the terminology "differentiable types" is used as a shorthand for "types that can be used as arguments and results of differentiable functions". This notion is important because not all types are "differentiable" in this sense. For example, types representing real numbers and vector spaces are "differentiable", but strings and integers are not.
```
rxwei:
```suggestion
Speaking in terms of elementary calculus, only functions are "differentiable": only functions have derivatives and can be differentiated. In this document, the terminology "differentiable types" is used as a shorthand for "types that can be used as arguments and results of differentiable functions". This notion is important because not all types are "differentiable" in this sense. For example, types representing real numbers and vector spaces are "differentiable", but strings and integers are not.
```

Mathematically speaking, spaces can be differentiable as well.

docs/DifferentiableTypes.md
bartchr808: I noticed most of `AutoDiff.swift` in the Swift repo uses this syntax. I'm guessing there is a minor TODO JIRA ticket to update the file?

rxwei: The story is rather complicated. The standard library uses "the swift.org code example style", which is not officially documented and used only within the compiler repo. All our documentation and library code (like the deep learning library) use the [Google Swift Style](https://google.github.io/swift/) with the exception that indentations are 4 spaces long.

docs/DifferentiableTypes.md
```diff
@@ -0,0 +1,289 @@
...
+
+Here’s an example deep learning layer with some `@noDerivative` properties:
+
+```swift
+struct DenseLayer : Differentiable {
+    // These properties should have derivative values.
+    var weight: Tensor<Float>
+    var bias: Tensor<Float>
+
+    // These auxiliary properties should not have derivative values.
```
rxwei: It seems that these comments are line-wrapped based on the 80-column limit, not 100-column.

docs/DifferentiableTypes.md
```diff
@@ -0,0 +1,289 @@
...
+              TangentVector.CotangentVector == CotangentVector,
+              TangentVector.AllDifferentiableVariables == TangentVector
+    /// The cotangent bundle of this differentiable manifold.
+    associatedtype CotangentVector : AdditiveArithmetic & Differentiable
+        where CotangentVector.TangentVector == CotangentVector,
+              CotangentVector.CotangentVector == TangentVector,
+              CotangentVector.AllDifferentiableVariables == CotangentVector
+    /// The type of all differentiable variables in this type.
+    associatedtype AllDifferentiableVariables : Differentiable
+        where AllDifferentiableVariables.AllDifferentiableVariables ==
```
rxwei: Under the 100-column limit, you can specify this whole constraint in one line.

docs/DifferentiableTypes.md
```diff
@@ -0,0 +1,289 @@
...
+    public typealias AllDifferentiableVariables = Tensor
+}
+```
+
+## Compiler-synthesized implementations
+
+As shown above, the compiler automatically synthesizes implementations of `Differentiable` requirements for struct types.
+
+Here are the current conditions for synthesis:
+* The type must declare a conformance to `Differentiable`, either on the type declaration or on an extension.
```
rxwei:
```suggestion
* The type must declare a conformance to `Differentiable`, either on the type declaration or on an extension in the same file as the declaration.
```

docs/DifferentiableTypes.md
```diff
@@ -0,0 +1,289 @@
...
+* The conforming type must be a struct type.
```
rxwei: Either "`struct`" or "structure type".
```suggestion
* The conforming type must be a `struct`.
```

docs/DifferentiableTypes.md
```diff
@@ -0,0 +1,289 @@
...
+* All stored properties of the conforming type must either conform to `Differentiable` or be marked with the `@noDerivative` attribute.
```
rxwei:
```suggestion
* All stored properties of the conforming type must either conform to `Differentiable` or be marked with the `@noDerivative` attribute. If a non-`Differentiable` stored property is not marked with `@noDerivative`, it will be treated as if it had `@noDerivative` and a compiler warning (with a fix-it in IDEs) will be produced to ask the user to make it explicit.
```

docs/DifferentiableTypes.md
```diff
@@ -0,0 +1,289 @@
...
+// Example when `AllDifferentiableVariables == Self`.
+var allDifferentiableVariables: AllDifferentiableVariables {
+    get { return self }
+    set { return newValue }
+}
+
+// Example when `AllDifferentiableVariables != Self`.
+var allDifferentiableVariables: AllDifferentiableVariables {
+    get { return AllDifferentiableVariables(x: x, y: y, ...) }
+    set { x = newValue.x; y = newValue.y }
```
rxwei: For consistency with `get`:
```suggestion
    set { x = newValue.x; y = newValue.y; ... }
```

docs/DifferentiableTypes.md
bartchr808: Ahh, I see makes sense thanks!

docs/DifferentiableTypes.md
```diff
@@ -0,0 +1,289 @@
...
+  * Regarding member struct synthesis: for each "differentiation property" in the conforming type, a corresponding stored property is synthesized in the member structs, with type equal to the property’s associated type.
+  * `TangentVector` and `CotangentVector` can be set to `AllDifferentiableVariables` if all differentiation properties satisfy `Self.TangentVector == Self.CotangentVector == Self.AllDifferentiableVariables`.
+
+A memberwise initializer is synthesized for the conforming type itself, in addition to all associated structs. This is important for differentiating struct properties accesses and synthesizing manifold operation requirements.
+
+### Synthesis for other requirements
+
+`var allDifferentiableVariables: AllDifferentiableVariables` is synthesized as a computed property that mirrors the differentiation properties of the conforming type.
+
+It is always synthesized with a getter.
```
rxwei: Indent these two rules with a bullet?

docs/DifferentiableTypes.md
dan-zheng: I don't believe this behavior is currently true, though we definitely want to enforce it.
Filed [TF-350](https://bugs.swift.org/browse/TF-350) to track it.

rxwei: I see. We should still document the correct rule that we want this to be, like `allDifferentiableVariables`'s setter.

docs/DifferentiableTypes.md
dan-zheng: Changed to abbreviated names, as suggested.

docs/DifferentiableTypes.md
dan-zheng: I provided more context for `Tensor` (stated that it's from the TensorFlow library) and linked to API docs. I think this is sufficient.

docs/DifferentiableTypes.md
dan-zheng: I mentioned SIMD vector types, but said "conditional conformances will be added to floating-point SIMD vector types", which is accurate.

Thanks for mentioning `Float80`: I added it.

docs/DifferentiableTypes.md
```diff
@@ -0,0 +1,293 @@
...
+  <br>
+  <sub>From https://en.wikipedia.org/wiki/Pushforward_(differential).</sub>
+</p>
+
+Here is a detailed explanation of the `Differentiable` protocol:
+* `associatedtype TangentVector` represents the type of directional derivatives computed via forward-mode differentiation.
+* `associatedtype CotangentVector` represents the type of gradient values computed via reverse-mode differentiation.
+  * `CotangentVector` types are used and produced by differential operators like `gradient` and `pullback`.
+* `var allDifferentiableVariables: AllDifferentiableVariables` represents all differentiable variables in an instance of the conforming type, where `associatedtype AllDifferentiableVariables` is the type of all differentiable variables.
+  * The motivation/design behind "all differentiable variables" is enabling key-path-based parameter optimization by making parameters and their gradients have the same type. Read the [synthesis rules](#compiler-synthesized-implementations) below and the [parameter optimization doc][parameter-optimization] for more information.
```
rxwei:
```suggestion
  * The motivation/design behind "all differentiable variables" is enabling key-path-based parameter optimization by making parameters and their gradients have the same type. Read the [synthesis rules](#compiler-synthesized-implementations) below and the [parameter optimization document][parameter-optimization] for more information.
```

docs/DifferentiableTypes.md
dan-zheng: Done! I agree, just now committed changes.

PR #145, opened 2019-03-15
Add differentiable functions doc.

dan-zheng:
Add "Differentiable functions and differentiation APIs" document, explaining:
- `@differentiable` function types.
- `@differentiable` declarations.
- differentiation APIs.

### DIFF
--- a/docs/DifferentiableFunctions.md
+++ b/docs/DifferentiableFunctions.md
@@ -1,0 +1,444 @@
+# Differentiable functions and differentiation APIs
+
+[Richard Wei], [Dan Zheng], [Marc Rasi], [Parker Schuh]
+
+Last updated: March 2019
+
+## Introduction
+
+Swift supports differentiable functions as part of the language. The
+`@differentiable` attribute appears in two locations in Swift syntax: as an
+annotation on function types and as an annotation on function declarations (and
+other similar declarations). This document explains the meaning of these
+annotations.
+
+## `@differentiable` function type attribute
+
+### Basics
+
+In Swift, function types can have attributes. When a function type is annotated
+with `@differentiable`, Swift guarantees that all values of that function type
+can be differentiated.
+
+`@differentiable` functions can be called like normal functions, or be passed to
+APIs that take `@differentiable` functions, like [`gradient(of:)`]. The binary
+representation of a `@differentiable` function is a special data structure
+containing the original function along with extra information required for
+computing its derivatives. Usage `@differentiable` functions are a part of
+Swift's type system. Most notably, they are used by differentiation APIs in the
+standard library. Here are some examples demonstrating differentiation APIs:
+
+```swift
+func square(_ x: Float) -> Float {
+    return x * x
+}
+let x: Float = 3.0
+
+// Free function examples.
+
+// Computes the gradient of `square` at `x`.
+print(gradient(at: x, in: square)) // 6.0
+// Computes the gradient of `square`, then applies it to `x`.
+print(gradient(of: square)(x)) // 6.0
+// Computes the value and gradient of `square` at `x`.
+print(valueWithGradient(at: x, in: square)) // (value: 9.0, gradient: 6.0)
+
+// Method examples.
+
+// Computes the gradient of `square` at `x`.
+print(x.gradient(in: square)) // 6.0
+// Computes the value and gradient of `square` at `x`.
+print(x.valueWithGradient(in: square)) // (value: 9.0, gradient: 6.0)
+```
+
+Here's a list of differentiation APIs provided by the standard library:
+
+| Differentiation APIs  | Description                                                                                                                                                      |
+|-----------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| [`valueWithPullback(at:in:)`] <br/> [`valueWithPullback(at:_:in:)`] | Returns original result and backpropagation function. <br/> **Important note: this is the core differentiation API. All other APIs are defined in terms of `valueWithPullback`.** |
+| [`pullback(at:in:)`] <br/> [`pullback(at:_:in:)`] | Returns backpropagation function. |
+| [`gradient(at:in:)`] <br/> [`gradient(at:_:in:)`] | Returns partial derivatives with respect to arguments. |
+| [`valueWithGradient(at:in:)`] <br/> [`valueWithGradient(at:_:in:)`] | Returns original result and partial derivatives with respect to arguments. |
+| [`gradient(of:)`] <br/> [`gradient(of:)` (arity 2)] | Returns gradient function. |
+| [`valueWithGradient(of:)`] <br/> [`valueWithGradient(of:)` (arity 2)] | Returns gradient function. |
+
+### Constructing `@differentiable` functions
+
+A value with a non-differentiable function type can be implicitly converted to
+one with a corresponding `@differentiable` function type. In fact, this is what
+happened in the example above:
+
+```swift
+// `square` has type `(Float) -> Float`.
+func square(_ x: Float) -> Float {
+    return x * x
+}
+let x: Float = 3.0
+
+// The second argument of `gradient(at:in:)` has type `@differentiable (Float) -> Float`.
+// When `square` is passed to `gradient(at:in:)`, it is implicitly converted to a value with the
+// `@differentiable` type.
+print(gradient(at: x, in: square)) // 6.0
+```
+
+The implicit conversion from a value with type `(T) -> U` to a value with type
+`@differentiable (T) -> U` actually triggers differentiation by the compiler.
+Thus, differentiation is type-driven.
+
+If differentiation succeeds, the `@differentiable (T) -> U` value is
+constructed. If differentiation fails, the compiler emits a compile-time error:
+
+```swift
+let add: @differentiable (Float, Float) -> Float = { x, y in
+    // The `Int` initializer call below is non-differentiable.
+    Float(Int(x + y))
+}
+```
+
+```console
+test.swift:1:52: error: function is not differentiable
+let add: @differentiable (Float, Float) -> Float = { x, y in
+                                                   ^~~~~~~~~
+test.swift:2:11: note: cannot differentiate through a non-differentiable result; do you want to add '.withoutDerivative()'?
+    Float(Int(x + y))
+          ^
+```
+
+There are a few reasons why differentiation can fail:
+* The function to differentiate contains non-differentiable computation along
+  the dataflow from parameters to function result.
+* The function to differentiate is defined in another module.
+* The function to differentiate uses [control
+  flow](https://docs.swift.org/swift-book/LanguageGuide/ControlFlow.html)
+  (if-statements, switch-statements, loops, etc). This restriction will be
+  lifted soon.
+  
+## `@differentiable` declaration attribute
+
+### Basics
+
+The `@differentiable` attribute can also be applied to function declarations.
+`@differentiable` marks a function as being differentiable with respect to some
+parameters (the varying parameters, explained below). `@differentiable` requires
+the types of the varying parameters and the function result type to all conform
+to the `Differentiable` protocol.
+
+This annotation does not change the declaration to have a `@differentiable`
+function type; instead, it triggers differentiation by the compiler on the
+function. If differentiation succeeds, then conversion of the function to a
+`@differentiable` function is guaranteed to succeed later.
+
+You may wonder about the purpose of the `@differentiable` declaration attribute,
+given that non-differentiable functions can implicitly be converted to
+`@differentiable` functions, as mentioned above. The main reason is that the
+`@differentiable` declaration attribute is a contract for differentiability: if
+a function is declared with `@differentiable` and it compiles, then it is always
+guaranteed to be differentiable, even in other modules. On the other hand, if a
+function is not declared with `@differentiable`, then differentiation of the
+function in other modules will fail.
+
+This is why floating-point operations in the standard library are declared with
+`@differentiable`:
+
+```swift
+extension Float {
+    @differentiable
+    public static func + (lhs: Float, rhs: Float) -> Float { ... }
+}
+```
+
+Besides function declarations, there are a few other function-like declarations
+that can be marked with `@differentiable`:
+- Computed property getters. (This requires both the type defining the property
+  and the type of the property to conform to `Differentiable`.)
+- Initializers. (This requires the type defining the initializer to conform to
+  `Differentiable`.)
+
+For instance methods defined on types that conform to `Differentiable`, the
+`self` property can be marked as a varying parameter. Derivatives of these
+methods return the partial derivative with respect to `self`. For these methods,
+`@differentiable` infers `self` as a varying parameter by default.
+
+```swift
+struct Vector: Differentiable, VectorNumeric {
+    var x, y: Float
+
+    // Differentiable computed property.
+    @differentiable // Implicitly: @differentiable(wrt: self)
+    var magnitude: Float {
+        return (x * x + y * y).squareRoot()
+    }
+
+    // Differentiable initializer.
+    @differentiable // Implicitly: @differentiable(wrt: (x, y))
+    init(x: Float, y: Float) {
+        self.x = x
+        self.y = y
+    }
+}
+
+let v = Vector(x: 2, y: 2)
+print(v.magnitude)
+// 2.828427
+print(gradient(at: v) { v in v.magnitude })
+// Vector(x: 64.0, y: 64.0)
+```
+
+### Differentiating with respect to
+
+Mathematically, let "varying parameters" refer to the parameters (i.e.
+independent variables) of a differentiable function whose partial derivatives
+are computed by the function's derivative.
+
+By default, the `@differentiable` attribute infers all function parameters that
+conform to `Differentiable` to be the varying parameters. However, this is not
+always desirable. To explicitly declare functions as differentiable with respect
+to a subset of parameters, explicitly specify the varying parameters using the
+`@differentiable(wrt: ...)` syntax.
+
+Here's an example of a 2-D convolution operation, adapted from the TensorFlow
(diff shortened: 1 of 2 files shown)


docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,444 @@
...
+let x: Float = 3.0
+
+// Free function examples.
+
+// Computes the gradient of `square` at `x`.
+print(gradient(at: x, in: square)) // 6.0
+// Computes the gradient of `square`, then applies it to `x`.
+print(gradient(of: square)(x)) // 6.0
+// Computes the value and gradient of `square` at `x`.
+print(valueWithGradient(at: x, in: square)) // (value: 9.0, gradient: 6.0)
```
dan-zheng: It may be nice to include examples of `pullback` and `valueWithPullback`.
Also, it may be nice to include examples of differentiation APIs for higher arity functions.

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,444 @@
...
+## `@differentiable` function type attribute
+
+### Basics
+
+In Swift, function types can have attributes. When a function type is annotated
+with `@differentiable`, Swift guarantees that all values of that function type
+can be differentiated.
+
+`@differentiable` functions can be called like normal functions, or be passed to
+APIs that take `@differentiable` functions, like [`gradient(of:)`]. The binary
```
bartchr808: The `gradient(of:)` seems to be linking to this 
https://tensorflow.devsite.corp.google.com/swift/api_docs/Functions#/s:10TensorFlow8gradient2of15CotangentVectorQzxcAA0A0Vyq_Gxc_tAA14DifferentiableRzAA0aB13FloatingPointR_r0_lF 
which seems inaccessible

bartchr808: This is true for the links to the differentiation APIs table as well

dan-zheng: Thanks for the catch! Fixed.

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,444 @@
...
+```
+
+There are a few reasons why differentiation can fail:
+* The function to differentiate contains non-differentiable computation along
+  the dataflow from parameters to function result.
+* The function to differentiate is defined in another module.
+* The function to differentiate uses [control
+  flow](https://docs.swift.org/swift-book/LanguageGuide/ControlFlow.html)
+  (if-statements, switch-statements, loops, etc). This restriction will be
+  lifted soon.
```
bartchr808: Can't wait to help out with this! 😄

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,444 @@
...
+```
+
+All other differentiation APIs are defined in terms of `valueWithPullback`.
+Here's an example for `gradient`:
+
+```swift
+// `gradient` returns the partial derivative with respect to varying parameters for scalar-result
+// functions. It simply returns `pullback(1)`.
+func fooGradient(_ x: Float) -> Float {
+    let pullback = foo_valueWithPullback(x).pullback
```
bartchr808:
```suggestion
    let pullback = fooValueWithPullback(x).pullback
```

rxwei:
Great progress, I’ll have some time to review this later today!

docs/DifferentiableFunctions.md
dan-zheng: Thanks! Fixed in https://github.com/tensorflow/swift/pull/145/commits/a5cd494b0ec9ad754b793eaebffd7417987820a4.

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,444 @@
...
+## Constructing a `@differentiable` function from a derivative
+
+Given a function and its derivative, it is possible to construct a
+`@differentiable` version of the function using the
+`differentiableFunction(from:)` helper function defined in the standard library.
+
+Here's an example:
+```swift
+let multiply: @differentiable (Float, Float) -> Float =
+    differentiableFunction(from: { x, y in (value: x * y, pullback: { v in v * y, v * x }))
```
rxwei: Syntax error on the pullback closure. Don’t think this will compile.

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,444 @@
...
+print(gradient(of: sillyExp)(3))
+// Taking 𝑒(2.7182817) to the power of 3.0!
+// 20.085535
+```
+
+## Constructing a `@differentiable` function from a derivative
+
+Given a function and its derivative, it is possible to construct a
+`@differentiable` version of the function using the
+`differentiableFunction(from:)` helper function defined in the standard library.
```
rxwei: Link to API docs.

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,444 @@
...
+test.swift:1:52: error: function is not differentiable
+let add: @differentiable (Float, Float) -> Float = { x, y in
+                                                   ^~~~~~~~~
+test.swift:2:11: note: cannot differentiate through a non-differentiable result; do you want to add '.withoutDerivative()'?
+    Float(Int(x + y))
+          ^
+```
+
+There are a few reasons why differentiation can fail:
+* The function to differentiate contains non-differentiable computation along
```
rxwei: Earlier Conal Elliott on Twitter made a good point about the distinction between “non-differentiable functions” (mathematically non differentiable functions) and “functions whose derivatives are not computable” (opaque functions). Since our language doesn’t know math, we don’t really know what functions are really nondifferentiable in the mathematical sense. So we’ve been using “cannot differentiate xxx” in our diagnostics instead of “xxx is nondifferentiable”. We should probably use the same wording here instead of “nondifferentiable”.

docs/DifferentiableFunctions.md
dan-zheng: Thanks for the catch! Fixed.

docs/DifferentiableFunctions.md
dan-zheng: Done.

docs/DifferentiableFunctions.md
dan-zheng: How about:
`* The function to differentiate contains dataflow, from parameters to result, that cannot be differentiated.`

rxwei: “Data flow” or “dataflow” is a compiler term IMO. The user understands “code” or “computation”.

dan-zheng: How about:
```
* The function to differentiate contains computation, from parameters to result, that cannot be differentiated.
```

And also, how about adding another point for opaque functions?
```
* The function to differentiate is opaque, i.e. it is a function parameter with a non-`@differentiable` function type.
```

rxwei: I think both these are good. Any inability to differentiate something always comes from encountering an opaque, non-@differentiable function.

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,447 @@
...
+func sillyExp(_ x: Float) -> Float {
+    let 𝑒 = Float(M_E)
+    print("Taking 𝑒(\(𝑒)) to the power of \(x)!")
+    return pow(𝑒, x)
+}
+
+@differentiating(sillyExp)
+func sillyDerivative(_ x: Float) -> (value: Float, pullback: (Float) -> Float) {
+    let y = sillyExp(x)
+    return (value: y, pullback: { _ in y })
```
rxwei: Pullback should be v * y.

dan-zheng: Fixed, thanks.

docs/DifferentiableFunctions.md
dan-zheng: Added both points.

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,449 @@
...
+}
+```
+
+All other differentiation APIs are defined in terms of `valueWithPullback`.
+Here's an example for `gradient`:
+
+```swift
+// `gradient` returns the partial derivative with respect to varying parameters for scalar-result
+// functions. It simply returns `pullback(1)`.
+func fooGradient(_ x: Float) -> Float {
```
rxwei: The fact that 1) this is named something other than `gradient` and 2) this is not a differentiation API is a bit misleading. I'd suggest showing a completely generic `gradient(of:)` implementation here, and provide a link to the standard library implementation.

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,449 @@
...
+```swift
+extension Float {
+    @differentiable
+    public static func + (lhs: Float, rhs: Float) -> Float { ... }
+}
+```
+
+Besides function declarations, there are a few other function-like declarations
+that can be marked with `@differentiable`:
+- Computed property getters. (This requires both the type defining the property
```
rxwei: It's not just computed properties--same for stored properties, though their getter is not user-definable. You don't have to mention "getter" except to mention that this is making the getter operation differentiable w.r.t. `self`.

dan-zheng: Good point. I changed it to the following (didn't mention setters):
```
- Stored and computed properties.
  - This requires both the type defining the property
    and the type of the property to conform to `Differentiable`.
  - Property getters are differentiable with respect to `self`.
```

docs/DifferentiableFunctions.md
dan-zheng: I see - showing a generic `gradient(of:)` implementation using `valueWithPullback` is indeed more insightful.

Done.

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,454 @@
...
+//   - bias: [1.0, 1.0]
+print(𝛁input)
+// [[2.0, 2.0]]
+```
+
+## Providing a custom derivative
+
+Use the `@differentiating` attribute to mark a function as a custom derivative
+for another function. This is useful for registering derivatives for primitive
+operations.
```
rxwei: "primitive operations" -> "functions that cannot be differentiated by traversing the function body, typically a primitive math operator in a library"

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,454 @@
...
+    }
+    return original
+}
+```
+
+## `@differentiable` functions and automatic differentiation
+
+Automatic differentiation is the technique used by the compiler to automatically
+compute function derivatives. This document does not go into detail about
+automatic differentiation - but with an understanding of `@differentiable`
```
rxwei:
```suggestion
the automatic differentiation algorithm - but with an understanding of `@differentiable`
```

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,454 @@
...
+    func derivative(_ x: T) -> (value: R, pullback: (R.CotangentVector) -> T.CotangentVector) {
+        return vjp(x)
+    }
+    return original
+}
+```
+
+## `@differentiable` functions and automatic differentiation
+
+Automatic differentiation is the technique used by the compiler to automatically
```
rxwei: Add link: https://en.wikipedia.org/wiki/Automatic_differentiation

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,454 @@
...
+compute function derivatives. This document does not go into detail about
+automatic differentiation - but with an understanding of `@differentiable`
+functions and differentiation APIs, one can get a glimpse of how automatic
+differentiation works.
```
rxwei: Perhaps add some lines to imply to our users that "automatic differentiation" is just an implementation detail in the compiler that enables derivative calculation, and that only differentiation APIs and `@differentiable` attributes are what's to be internalized.

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,454 @@
...
+
+The key differentiation API is the [`valueWithPullback`] function, which takes a
+`@differentiable` function and arguments and returns two things: the result of
+the function when applied to arguments, and a backpropagation function called a
+"pullback", which takes the gradient of the result and returns the gradient of
```
rxwei: Reword:
"... which takes a direction vector (usually the partial derivative w.r.t. the original result) and returns the directional gradient w.r.t. the varying parameters.

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,454 @@
...
+
+## `@differentiable` functions and automatic differentiation
+
+Automatic differentiation is the technique used by the compiler to automatically
+compute function derivatives. This document does not go into detail about
+automatic differentiation - but with an understanding of `@differentiable`
+functions and differentiation APIs, one can get a glimpse of how automatic
+differentiation works.
+
+The key differentiation API is the [`valueWithPullback`] function, which takes a
```
rxwei:
```suggestion
The key differentiation API is the [`valueWithPullback(at:in:)`] function, which takes a
```

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,454 @@
...
+
+```swift
+func foo(_ x: Float) -> Float {
+    let double = x + x
+    let result = double * double
+    return result
+}
+```
+
+Conceptually, here's how the compiler computes `valueWithPullback` for `foo`:
```
rxwei:
```suggestion
Conceptually, here's how the compiler computes the `value` and the `pullback` for `foo`:
```

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,454 @@
...
+All other differentiation APIs are defined in terms of `valueWithPullback`.
+For example, here's how [`gradient(at:in:)`] is defined:
+
+```swift
+// `gradient` returns the partial derivative with respect to varying parameters for scalar-result
+// functions. It simply returns `pullback(1)`.
+func gradient<T, R>(
+    at x: T, in f: @differentiable (T) -> R
+) -> T.CotangentVector
+    where T : Differentiable, R : FloatingPoint & Differentiable,
```
rxwei:
```suggestion
    where T: Differentiable, R: FloatingPoint & Differentiable,
```

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,454 @@
...
+          R.CotangentVector == R {
+    let (value, pullback) = valueWithPullback(at: x, in: f)
+    return pullback(R(1))
+}
+
+print(gradient(at: x, in: foo)) // 24.0
+```
```
rxwei: Add a line below or above: 
"You can find [its implementation](https://github.com/apple/swift/blob/6e20cf1d2746435d879ccfeee9713aa7a32e1081/stdlib/public/core/AutoDiff.swift#L467) in the standard library."

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,454 @@
...
+### Basics
+
+The `@differentiable` attribute can also be applied to function declarations.
+`@differentiable` marks a function as being differentiable with respect to some
+parameters (the varying parameters, explained below). `@differentiable` requires
+the types of the varying parameters and the function result type to all conform
+to the `Differentiable` protocol.
+
+This annotation does not change the declaration to have a `@differentiable`
+function type; instead, it triggers differentiation by the compiler on the
```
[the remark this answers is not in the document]
dan-zheng: Changing to:
```
This annotation does not change the function declaration to have a
`@differentiable` function type; instead, it tells the compiler to differentiate
the function when compiling its containing file.
```

docs/DifferentiableFunctions.md
dan-zheng: Changing to:
```
Use the `@differentiating` attribute to mark a function as a custom derivative
for another function. This is useful for registering derivatives for functions
that cannot be differentiated by traversing the function body, typically
primitive math library operators.
```

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,454 @@
...
+
+```swift
+@differentiable(wrt: (input, filter))
+func conv2d(input: Tensor<Float>, filter: Tensor<Float>, strides: (Int, Int), padding: Padding) {
+    ...
+}
+```
+
+Functions can have multiple `@differentiable` attributes with differentiable
+`wrt` parameter lists. `@differentiable` protocol requirements If a protocol
```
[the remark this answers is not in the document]
dan-zheng: Oops, thanks!

docs/DifferentiableFunctions.md
dan-zheng: Adding:
```markdown
You can find [the implementation of
`gradient(at:in:)`](https://github.com/apple/swift/blob/6e20cf1d2746435d879ccfeee9713aa7a32e1081/stdlib/public/core/AutoDiff.swift#L467)
in the standard library.
```

docs/DifferentiableFunctions.md
dan-zheng: That's much better, thanks!

docs/DifferentiableFunctions.md
dan-zheng: Adding:
```markdown
*Note: automatic differentiation is an implementation detail in the compiler
that enables differentiation. `@differentiable` functions and differentiation
APIs are the important, visible features to understand.*
```

dan-zheng:
Thanks @rxwei for the thorough review!
All comments have addressed, ready for re-review.

docs/DifferentiableFunctions.md
```diff
@@ -0,0 +1,468 @@
...
+```
+
+## Providing a custom derivative
+
+Use the `@differentiating` attribute to mark a function as a custom derivative
+for another function. This is useful for registering derivatives for functions
+that cannot be differentiated by traversing the function body, typically
+primitive math library operators.
+
+Note: currently, the `@differentiating` attribute can only be used to define
```
rxwei: The other note paragraph is italic. Do you want to unify the style?

dan-zheng: Done: italicized both notes.

rxwei:
I think the document is good to go!

PR #148, opened 2019-03-17
Clarify documentation.

dan-zheng:


docs/DifferentiableFunctions.md
```diff
@@ -109,7 +109,8 @@ There are a few reasons why differentiation can fail:
   that cannot be differentiated.
 * The function to differentiate is opaque, i.e. it is a function parameter with
   a non-`@differentiable` function type.
-* The function to differentiate is defined in another module.
+* The function to differentiate is defined in another module, and both is not
```
rxwei: “Both is not ... does not ...” sounds a bit odd to me. 

Maybe reword the sentence to use “is neither ... nor has ...”?

dan-zheng: That's much better, thanks! Updated.

PR #155, opened 2019-03-20
Fix model_training_walkthrough.ipynb colab example url.

VolodymyrPavliukevych:
Just fixed file path.


saeta:
Thank you very much @VolodymyrPavliukevych ! Merging presently.

PR #157, opened 2019-03-21
Fix "A Swift Tour" tutorial.

dan-zheng:
The old version was outdated and had formatting errors.
- Update tutorial to match current contents of
  https://docs.swift.org/swift-book/GuidedTour/GuidedTour.html.
- Change "Experiment" sections to be code cells rather than text.
  Notebook viewers can enter code and experiment.

View the tutorial here: https://github.com/tensorflow/swift/blob/tutorials/docs/site/tutorials/a_swift_tour.ipynb

### DIFF
--- a/docs/site/tutorials/a_swift_tour.ipynb
+++ b/docs/site/tutorials/a_swift_tour.ipynb
@@ -61,7 +61,7 @@
       "metadata": {
         "id": "iNHY9-pzdE13",
         "colab_type": "code",
-        "outputId": "e1c49539-7b53-437f-f5cb-3a1099816d13",
+        "outputId": "8319f4a0-0f0c-440a-d547-89c4c30e205d",
         "colab": {
           "base_uri": "https://localhost:8080/",
           "height": 34
@@ -71,7 +71,7 @@
       "source": [
         "print(\"Hello, world!\")"
       ],
-      "execution_count": 0,
+      "execution_count": 1,
       "outputs": [
         {
           "output_type": "stream",
@@ -101,7 +101,7 @@
       },
       "cell_type": "markdown",
       "source": [
-        "### Simple Values\n",
+        "## Simple Values\n",
         "\n",
         "Use `let` to make a constant and `var` to make a variable. The value of a constant doesn’t need to be known at compile time, but you must assign it a value exactly once. This means you can use constants to name a value that you determine once but use in many places."
       ]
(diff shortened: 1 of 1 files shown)


dan-zheng:
View the tutorial here: https://github.com/tensorflow/swift/blob/tutorials/docs/site/tutorials/a_swift_tour.ipynb

docs/site/tutorials/a_swift_tour.ipynb
```diff
@@ -178,7 +188,7 @@
         "let width = 94\n",
         "label + String(width)"
       ],
-      "execution_count": 0,
+      "execution_count": 5,
```
saeta: Did you want to clear execution outputs before checking this in? Or is it intentional to have the outputs preserved?

docs/site/tutorials/a_swift_tour.ipynb
```diff
@@ -272,23 +292,82 @@
...
+          "base_uri": "https://localhost:8080/",
+          "height": 54
+        }
+      },
+      "cell_type": "code",
+      "source": [
+        "let quotation = \"\"\"\n",
+        "    Even though there's whitespace to the left,\n",
+        "    the actual lines aren't indented.\n",
+        "    Except for this line.\n",
```
saeta: I'm not sure I understand this line. Looking at the output of the cell, it doesn't look indented.

docs/site/tutorials/a_swift_tour.ipynb
```diff
@@ -842,21 +999,23 @@
...
     {
       "metadata": {
-        "id": "aVAMxeTxdE2v",
-        "colab_type": "text"
+        "id": "sgYex6gigH87",
+        "colab_type": "code",
+        "colab": {}
       },
-      "cell_type": "markdown",
+      "cell_type": "code",
```
saeta: Since this cell is just comments, does it make sense for this to be converted back to markdown? Or is there another reason for it to be a code cell?

(Note: this pattern of changes is repeated in other cells too.)

docs/site/tutorials/a_swift_tour.ipynb
lamberta: Typically, we don't save outputs. But Swift notebooks must save outputs because our tool that imports to tensorflow.org doesn't yet support swift notebook execution.
For that, we need a docker image with a Swift kernel that we can run `nbconvert` within.

docs/site/tutorials/a_swift_tour.ipynb
dan-zheng: Multiline strings in Swift as aligned based on the indentation of the closing triple quotes.

https://docs.swift.org/swift-book/LanguageGuide/StringsAndCharacters.html

docs/site/tutorials/a_swift_tour.ipynb
dan-zheng: "A Swift Tour" has experiment sections, which are exercises for the reader.

Example:
image

The original "A Swift Tour" is a static page, so everything shows up as text.
Since we're porting it to an interactive notebook, I think it makes sense to translate these into code sections. Users can actually code and experiment!

image

saeta:
Thanks all!

PR #163, opened 2019-03-25
Example Fix

rexmas:


rexmas:
I've signed the CLA and used the correct email/github user information. Not sure why the check is still in a failed state...

PR #179, opened 2019-04-29
reworked iris tutorial for latest api changes

brettkoonce:
I tried to do this cleanly, but it looks like juypter re-saved everything.  I can part it out and do the update line by line by hand so that the original notebook works the same, would you prefer that?

### DIFF
--- a/docs/site/tutorials/model_training_walkthrough.ipynb
+++ b/docs/site/tutorials/model_training_walkthrough.ipynb
@@ -1,1416 +1,1196 @@
 {
-  "nbformat": 4,
-  "nbformat_minor": 0,
-  "metadata": {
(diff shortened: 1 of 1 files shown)


rxwei:
Thanks a ton for your contributions! Actually, we already have a `tensorflow-0.3` branch which contains changes to all tutorials with API updates. Here's the updated Iris tutorial: https://github.com/tensorflow/swift/blob/tensorflow-0.3/docs/site/tutorials/model_training_walkthrough.ipynb.

Sorry for letting you know of duplicate work, and we'll communicate that to the community more clearly next time by opening an issue before working on it!

brettkoonce:
@rxwei no worries, should have checked better!

PR #180, opened 2019-04-29
Update the custom differentiation tutorial for v0.3.

saeta:


rxwei:
I think Marc has done it already: https://github.com/tensorflow/swift/blob/tensorflow-0.3/docs/site/tutorials/custom_differentiation.ipynb.

rxwei:
We'll just need to merge `tensorflow-0.3` once Colab is ready.

PR #190, opened 2019-05-07
tweak code samples to match latest api

brettkoonce:


docs/DesignOverview.md
```diff
@@ -119,14 +119,13 @@ Here are a couple of simple examples showing `Tensor` and `ShapedArray`:
...
 // `Tensor` examples.
-var matrix = Tensor<Float>(shape: [2, 2], scalars: [1, 2, 3, 4])
 var matrix: Tensor<Float> = [[1, 2], [3, 4]]
 // `matrix` represents [[1, 2], [3, 4]].
 
 // Arithmetic operations, using TensorFlow.
 let sum = matrix + matrix
-let sqrt = sqrt(matrix)
-let matrixProduct = matrix.dot(matrix)
+let sqrt = Raw.sqrt(matrix)
```
rxwei: This one didn't work because `sqrt` is colliding with the variable name. You can rename `let sqrt` to `let root` or something. Using `Raw` here is not recommended because there is a native Swift API for it, which is `sqrt(_:)`.

brettkoonce: kk, tweaked!

docs/DifferentiableFunctions.md
```diff
@@ -310,7 +310,7 @@ discussion](https://forums.swift.org/t/help-needed-with-retroactive-differentiab
 for more information.*
 
 ```swift
-import Darwin
+import Foundation
```
brettkoonce: @rxwei is this correct as well?

rxwei: This one is fine. `Foundation` transparently imports `Darwin`/`Glibc` based on the operating system.

rxwei:
Thanks!

PR #193, opened 2019-05-21
Update more references for Layer API in tutorials 

lukereichold:
Both the custom differentiation and model training walkthrough tutorials contain references to the old (pre- v0.3) `applied(to:)` on `Layer`. This updates these remaining references to instead use the new `call(_:)` method, reflecting the current docs.

### DIFF
--- a/docs/site/tutorials/custom_differentiation.ipynb
+++ b/docs/site/tutorials/custom_differentiation.ipynb
@@ -469,7 +469,9 @@
    "source": [
     "## Recomputing activations during backpropagation to save memory (checkpointing)\n",
     "\n",
-    "Checkpointing is a traditional technique in reverse-mode automatic differentiation to save memory when computing derivatives by making large intermediate values in the original computation not be saved in memory for backpropagation, but instead recomputed as needed during backpropagation. This technique has been realized in modern deep learning libraries as well. In Swift, API [`withRecomputationInPullbacks(_:)`](https://www.tensorflow.org/swift/api_docs/Protocols/Differentiable#/s:10TensorFlow14DifferentiablePAAE28withRecomputationInPullbacksyqd__qd__xcAaBRd__lF) enables you to control what to recompute during backpropagation, and it is available on all `Differentiable` types.\n",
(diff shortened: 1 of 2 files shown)


docs/site/tutorials/custom_differentiation.ipynb
```diff
@@ -507,9 +507,9 @@
    "outputs": [],
    "source": [
     "/// Given a differentiable function, returns the same differentiable function except when\n",
-    "/// derivatives of this function is being computed, values in the original function that are needed\n",
-    "/// for computing the derivatives will be recomputed, instead of being captured by the differential\n",
-    "/// or pullback.\n",
+    "/// derivatives of this function are being computed. In this case, values in the original function \n",
```
dan-zheng: Minor suggestions:
```
In that case, values in the original function needed for computing the derivatives will be recomputed, instead of being captured by the differential or pullback.
```

docs/site/tutorials/custom_differentiation.ipynb
```diff
@@ -469,7 +469,7 @@
...
+    "Checkpointing is a traditional technique in reverse-mode automatic differentiation to save memory when computing derivatives by making large intermediate values in the original computation not be saved in memory for backpropagation, but instead recomputed as needed during backpropagation. This technique has been realized in modern deep learning libraries as well. In Swift, API [`withRecomputationInPullbacks(_:)`](https://www.tensorflow.org/swift/api_docs/Protocols/Differentiable#/s:10TensorFlow14DifferentiablePAAE28withRecomputationInPullbacksyqd__qd__xcAaBRd__lF) enables you to control what to recompute during backpropagation, and it is available on all `Differentiable` types.\n",
```
dan-zheng: I might suggesting breaking up the first sentence:
```
Checkpointing is a traditional technique in reverse-mode automatic differentiation for saving memory.
Rather than saving large intermediate values in the original computation for computing derivatives,
the intermediate values are instead recomputed as needed during backpropagation.
```

lukereichold: Good points, thanks. Fixed.

PR #218, opened 2019-07-24
Add type Double explanation

8bitmp3:
Small suggestion: add a short explanation for why type `Double` was used in a floating point number example. This is because one can use either `Float` or `Double`. Can also emphasize the number of decimal points. Cheers

### DIFF
--- a/docs/site/tutorials/a_swift_tour.ipynb
+++ b/docs/site/tutorials/a_swift_tour.ipynb
@@ -130,7 +130,7 @@
       "source": [
         "A constant or variable must have the same type as the value you want to assign to it. However, you don’t always have to write the type explicitly. Providing a value when you create a constant or variable lets the compiler infer its type. In the example above, the compiler infers that `myVariable` is an integer because its initial value is an integer.\n",
         "\n",
-        "If the initial value doesn’t provide enough information (or if there is no initial value), specify the type by writing it after the variable, separated by a colon. Note that using type `Double` instead of `Float` for floating point numbers gives you more precision."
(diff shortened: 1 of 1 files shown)


docs/site/tutorials/a_swift_tour.ipynb
```diff
@@ -130,7 +130,7 @@
       "source": [
         "A constant or variable must have the same type as the value you want to assign to it. However, you don’t always have to write the type explicitly. Providing a value when you create a constant or variable lets the compiler infer its type. In the example above, the compiler infers that `myVariable` is an integer because its initial value is an integer.\n",
         "\n",
-        "If the initial value doesn’t provide enough information (or if there is no initial value), specify the type by writing it after the variable, separated by a colon."
+        "If the initial value doesn’t provide enough information (or if there is no initial value), specify the type by writing it after the variable, separated by a colon. Note that using type `Double` instead of `Float` for floating point numbers gives you more precision."
```
saeta:
```suggestion
        "If the initial value doesn’t provide enough information (or if there is no initial value), specify the type by writing it after the variable, separated by a colon. Note: using `Double` instead of `Float` for floating point numbers gives you more precision, and is the default floating point type in Swift."
```

8bitmp3:
Thanks @saeta, all done—just a CLA thing given that we have different email domains or something.

Cool cool. @rxwei I am doing a visual diff check between the two Swift tours and so far so good. The opaque types and other 5.1 stuff may be introduced in the Swift Tour by the folks at Apple maybe when 5.1 is out of dev? I understand that S4TF (Colab version at least) uses the `5.1-dev` build.

saeta:
This LGTM. Thanks!

PR #241, opened 2019-08-21
Assorted improvements to libSIL

apaszke-g:
Might be a good idea to land without squashing, because they're all independent.

cc @burmako

### DIFF
--- a/Sources/SIL/SIL.swift
+++ b/Sources/SIL/SIL.swift
@@ -290,6 +290,7 @@ public enum Enforcement {
 public enum FunctionAttribute {
     case alwaysInline
     case differentiable(_ spec: String)
+    case dynamicallyReplacable
     case noInline
     case semantics(_ value: String)
     case serialized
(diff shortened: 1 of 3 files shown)


Sources/SIL/SIL.swift
```diff
@@ -10,6 +10,11 @@ public class Module {
         let parser = try SILParser(forPath: silPath)
         return try parser.parseModule()
     }
+
+    public static func parse(fromString silString: String) throws -> Module {
```
rxwei: Per Swift API Design Guidelines, argument labels should not repeat the type information.
```suggestions
public static func parse(from silString: String) throws -> Module {
```

Sources/SIL/SIL.swift
```diff
@@ -290,7 +295,9 @@ public enum Enforcement {
 public enum FunctionAttribute {
     case alwaysInline
     case differentiable(_ spec: String)
+    case dynamicallyReplacable
     case noInline
+    case readonly
     case semantics(_ value: String)
```
rxwei: Enum payloads don't need argument names or labels. 

`case semantics(String)` is totally fine.

Sources/SIL/SIL.swift
```diff
@@ -355,7 +368,7 @@ public enum TupleElements {
 }
 
 // Reverse-engineered from -emit-sil
-public indirect enum Type {
+public indirect enum Type: Equatable {
     case addressType(_ type: Type)
```
rxwei: * Enum case names don't need to repeat the enum type information. I'd suggest dropping the "Type" suffixes.
* No need for argument names.

`address(Type)`

PR #242, opened 2019-08-21
Store blobs as [UInt8]

apaszke-g:


Sources/SIL/Bitcode.swift
```diff
@@ -14,7 +14,7 @@ indirect enum OperandKind {
 
 indirect enum BitcodeOperand {
     case bits(_ value: Bits)
-    case blob(_ value: Bits)
+    case blob(_ value: [UInt8])
```
rxwei: `blob([UInt8])`

Sources/SIL/Bitstream.swift
```diff
@@ -7,6 +7,7 @@ struct Bits: Equatable, Hashable, ExpressibleByIntegerLiteral, CustomStringConve
     private var bits: [Bool]
     var description: String { String(bits.reversed().map { $0 ? "1" : "0" }) }
     var count: Int { bits.lastIndex(of: true).map { $0 + 1 } ?? 0 }
+    var isZero: Bool { bits.allSatisfy { !$0 } }
```
rxwei: Randomly:
```suggestion
    var isZero: Bool { bits.allSatisfy(!) }
```

Sources/SIL/Bitstream.swift
```diff
@@ -45,15 +46,6 @@ struct Bits: Equatable, Hashable, ExpressibleByIntegerLiteral, CustomStringConve
         return Int(asUInt32())
     }
 
-    func asString() -> String? {
```
rxwei: "asXXX" is not a common pattern in Swift APIs. Usually we'd define a computed property with a noun as the name.

eaplatanios: You can also use the [`CustomStringConvertible`](https://developer.apple.com/documentation/swift/customstringconvertible) for this particular use case with a field called `description`.

PR #243, opened 2019-08-23
Make sure that libSIL tests run on gLinux

burmako:
1) Explicitly defines test cases to run instead of relying on macOS-specific magic to autodetect them. This is mostly straightforward albeit tedious almost everywhere, except InstructionTests which need special treatment.

2) Instead of having test data built every time when tests are run by whatever `swiftc` installed on the current machine, the pull requests uses newly introduced predefined test data produced by swift-DEVELOPMENT-SNAPSHOT-2019-08-22-a. This is motivated by the fact that we cannot expect the same version of `swiftc` to be installed everywhere we will be running tests (notably, in Google3 presubmits we cannot expect any `swiftc` in `PATH`).

3) Disables testLoadingSIB. I could never get this test working on my laptop, and I've run into the same problem on my gLinux desktop now. Since this functionality is not on our critical paths, I've disabled the test and made a note to fix it later (https://bugs.swift.org/browse/TF-774).


Tests/SILTests/BitcodeTests.swift
```diff
@@ -1,52 +1,45 @@
...
-          <SIL_ONE_VALUE_ONE_OPERAND op0=88 op1=0 op2=1 op3=6 op4=0 op5=2/>
-          <SIL_INST_APPLY op0=2 op1=0 op2=7 op3=0 op4=8 op5=4 op6=7 op7=0 op8=3 op9=7 op10=0/>
-          <SIL_ONE_TYPE_VALUES op0=87 op1=6 op2=0 op3=7 op4=0 op5=5/>
-          <SIL_ONE_OPERAND op0=113 op1=0 op2=6 op3=0 op4=6/>
-        </SIL_BLOCK>
-
-        """
-            )
-        }
+        // TODO(TF-774): Disabling the test since it's not on the critical path at the moment.
```
apaszke-g: Instead of commenting everything, can you either:
1. Just put a `return` on the first line of this function
2. Put it in a separate function and decorate with this comment?

I'll try to enable this test again soon, so let's not avoid making the diffs super large for now.

Tests/SILTests/InstructionTests.swift
```diff
@@ -1,90 +1,113 @@
...
+#if os(Linux)
+    public final class InstructionTests: XCTestCase {
+    }
+
+    extension InstructionTests {
+        public static var allTests = instructionDefs.map { instructionDef in
+            (
+                "testRoundtrip",
+                { (_: InstructionTests) in
+                    {
```
apaszke-g: Man, this whitespace 😮

Tests/SILTests/InstructionTests.swift
```diff
@@ -1,90 +1,113 @@
...
+#endif
+
+#if os(Linux)
+    public final class InstructionTests: XCTestCase {
+    }
+
+    extension InstructionTests {
+        public static var allTests = instructionDefs.map { instructionDef in
+            (
+                "testRoundtrip",
```
apaszke-g: Can all of those tests really have the same name?

Tests/SILTests/InstructionTests.swift
```diff
@@ -1,90 +1,113 @@
...
+                    testSuite.addTest(testCase)
+                }
             }
+            return testSuite
+        }
+    }
+#endif
+
+#if os(Linux)
+    public final class InstructionTests: XCTestCase {
```
apaszke-g: Do we really need both paths? Doesn't this one work on macOS?

Tests/SILTests/ModuleTests.swift
```diff
@@ -2,31 +2,24 @@ import Foundation
...
-                }
-            } catch {
-                XCTFail(String(describing: error))
+    private func testRoundtrip(_ silPath: String) {
+        do {
+            let expected = stripUnsupportedSILSyntax(silPath)
+            let module = try Module.parse(fromSILPath: silPath)
+            let actual = module.description + "\n"
+            if (expected != actual) {
+                let actualPath = FileManager.default.makeTemporaryFile()!.path
```
apaszke-g: nit: handle the failure in a way that's nicer than a force unwrap

Tests/SILTests/InstructionTests.swift
burmako: Courtesy of swift-format :)

Tests/SILTests/InstructionTests.swift
burmako: This how it works on macOS. Apparently they can 🤷‍♂️

Tests/SILTests/InstructionTests.swift
burmako: It doesn't because allTests are only used in LinuxMain.swift.

Tests/SILTests/BitcodeTests.swift
burmako: Done.

Tests/SILTests/ModuleTests.swift
burmako: Done.

PR #261, opened 2019-08-29
Update README logo.

dan-zheng:
- Use TensorFlow 2.0 logo.
- Use transparent instead of white background.
- Remove plus symbol.

Check out the logo [here](https://raw.githubusercontent.com/dan-zheng/tensorflow-swift/logo/images/logo.png) and the README [here](https://github.com/dan-zheng/tensorflow-swift/blob/logo/README.md).


dan-zheng:
> The new size really doesn't look nice because it's too large and broke the original near golden-ratio placement: the original ratio between the width of visible logo and the visual text width. For optimizing visual perception, please make sure the new logo matches the exact visible width of the old logo.

Sorry, that was a WIP commit, not ready for review.
I've reverted it, please ignore that overly tall image.

dan-zheng:
Current status: the logo is high resolution (large) and manually scaled down using the HTML `height` attribute.

Unfortunately, this doesn't preserve aspect ratio on small screens:



---

Possible solutions:
* Recommend: use a smaller logo image (that's sufficiently high-res) with the exact desired height and don't resize it. This is how the current logo works. I tried to [downsize the new logo in this patch but it became too low-res](https://github.com/dan-zheng/tensorflow-swift/blob/logo-wip/images/logo-small.png). 
* Find some other way to resize large images while preserving aspect ratio. I'm not sure this is possible with GitHub Markdown because `style` attributes are erased.

@markdaoust: I believe you helped us create the original Swift for TensorFlow logo. I wonder if you could please help create a new one, updated with the TF 2.0 logo?

MarkDaoust:
I didn't make the original. Would switcing to svg solve some of these problems?

https://upload.wikimedia.org/wikipedia/commons/2/20/Swift_logo_with_text.svg
https://drive.google.com/open?id=1URqDzFmnPFOQmVXWqEoMGVho3J-_6k3k

dan-zheng:
> I didn't make the original. Would switcing to svg solve some of these problems?
> 
> https://upload.wikimedia.org/wikipedia/commons/2/20/Swift_logo_with_text.svg
> https://drive.google.com/open?id=1URqDzFmnPFOQmVXWqEoMGVho3J-_6k3k

Thanks for sharing! SVG would be awesome - I don't have photo editing software to stitch together the two SVGs though.

dan-zheng:
Closing in favor of https://github.com/tensorflow/swift/pull/302.

PR #263, opened 2019-09-05
Add support for partial_apply and convert_escape_to_noescape

apaszke-g:


Sources/SIL/SILParser.swift
```diff
@@ -182,11 +189,23 @@ class SILParser: Parser {
             let value = try parseInt()
             return .integerLiteral(type, value)
         case "load":
+            let _ = skip("[copy]")
+            let _ = skip("[take]")
+            let _ = skip("[trivial]")
```
burmako: Let's have these as fields on `load` (and add a test that uses all these flags)?

Sources/SIL/SILParser.swift
```diff
@@ -199,6 +218,7 @@ class SILParser: Parser {
         case "store":
             let value = try parseValue()
             try take("to")
+            let _ = skip("[init]") // Used in ownership SSA
             let _ = skip("[trivial]") // Used in ownership SSA
```
burmako: Likewise, let's add these flags to fields of `store`.

Sources/SIL/SILParser.swift
```diff
@@ -610,7 +630,7 @@ class SILParser: Parser {
     func parseTypeName() throws -> String {
         let start = position
         // TODO(#14): Make name parsing more thorough.
-        let name = take(while: { $0.isLetter || $0.isNumber || $0 == "_" })
+        let name = take(while: { $0.isLetter || $0.isNumber || "_[]".contains($0) })
```
burmako: When is this necessary? Let's add a test?

Sources/SIL/SILPrinter.swift
```diff
@@ -163,6 +170,15 @@ class SILPrinter: Printer {
...
+            print(when: calleeGuaranteed, "[callee_guaranteed] ")
+            print(when: onStack, "[on_stack] ")
+            print(value)
+            print(whenEmpty: false, "<", substitutions, ", ", ">") { naked($0) }
+            print("(", arguments, ", ", ")") { print($0) }
+            print(" : ")
+            print(type)
         case let .pointerToAddress(operand, strict, type):
             print("pointer_to_address ")
             print(operand)
```
burmako: Am I missing something or this PR doesn't have instruction tests?

Sources/SIL/SILParser.swift
apaszke-g: Hmm I didn't want to pollute the instruction with too many fields, and I don't know whether they're exclusive or not, so I'm not sure whether we can make that an enum too. This stuff is entirely undocumented and so I ignored it. I can add it though if you really want it.

Sources/SIL/SILParser.swift
apaszke-g: Apparently SILGen might contain desugared types 😕

Sources/SIL/SILPrinter.swift
apaszke-g: Oh I must have forgotten to add those. Will do.

burmako:
Could you give an example?

Sources/SIL/SILParser.swift
apaszke-g: Here you go:
```
%149 = apply %148<[Int], PartialRangeFrom<Int>>(%143, %146, %144)
```

apaszke-g: Hmmm now that I think about it, it might be better to normalize `[T]` to `Array<T>`

burmako:
I think it would be could to update the parser to treat [T] as a special
kind of type instead of stuffing it into a name.

apaszke-g:
Yep, I've implemented the desugaring in the parser. That was an ugly hack I made to get my stuff to work (I don't need those specializations), so thanks for catching this! I think all the comments should have been addressed by my last commit.

Sources/SIL/SILParser.swift
```diff
@@ -610,7 +643,7 @@ class SILParser: Parser {
     func parseTypeName() throws -> String {
         let start = position
         // TODO(#14): Make name parsing more thorough.
-        let name = take(while: { $0.isLetter || $0.isNumber || $0 == "_" })
+        let name = take(while: { $0.isLetter || $0.isNumber || "_".contains($0) })
```
burmako: Let's change this back to `$0 == "_"`? This reads a bit strangely.

Sources/SIL/SILPrinter.swift
```diff
@@ -157,12 +164,25 @@ class SILPrinter: Printer {
             print(type)
             print(", ")
             literal(value)
-        case let .load(operand):
+        case let .load(maybeOwnership, operand):
             print("load ")
+            if let ownership = maybeOwnership {
+                print(ownership)
+                print(" ")
+            }
```
burmako: `print(maybeOwnership, " ")`?

Sources/SIL/SILPrinter.swift
```diff
@@ -172,10 +192,14 @@ class SILPrinter: Printer {
...
             print(operand)
-        case let .store(value, operand):
+        case let .store(value, maybeOwnership, operand):
             print("store ")
             print(value)
             print(" to ")
+            if let ownership = maybeOwnership {
+                print(ownership)
+                print(" ")
+            }
```
burmako: Ditto.

Sources/SIL/SILPrinter.swift
apaszke-g: Ok, if we really want to code golf this, then I just realized that the current pattern for handling cases like this is
```
print(instructionDef.result, " = ") { print($0) }
```
which seems _really_ awkward to me. I can change it to that if you feel strongly about it though.

Sources/SIL/SILParser.swift
apaszke-g: sgtm

Sources/SIL/SILPrinter.swift
apaszke-g: We also use the `if` pattern [here (twice)](https://github.com/tensorflow/swift/blob/master/Sources/SIL/SILPrinter.swift#L322-L330) and [here](https://github.com/tensorflow/swift/blob/master/Sources/SIL/SILPrinter.swift#L135-L138).

PR #264, opened 2019-09-05
Emit unknown instruction upon parse failures

apaszke-g:
SILGen sometimes emit weird things, but I don't really care about them, and this lets the tool handle failures more gracefully.


apaszke-g:
So you want me to effectively put a malformed instruction in the unit tests? In general this code should only get triggered when e.g. `swiftc` changes the canonical SIL format or when we have a bug in our parsing code.

apaszke-g:
I've added the test so hopefully this should be good to land nwo

apaszke-g:
All tests passing locally.

PR #271, opened 2019-09-10
Fix debug location parsing

apaszke-g:
The old code would incorrectly consume commas that might belong
to other fields and it also expected the format presented in the SIL
documentation, which doesn't match what the compiler outputs today.


Sources/SIL/SILParser.swift
```diff
@@ -580,18 +583,24 @@ class SILParser: Parser {
...
+        //     not true once you look at the compiler outputs or its source code.
+        guard skip(",") else { return nil }
         let loc = try parseLoc()
-        guard scopeRef != nil || loc != nil else { return nil }
+        // NB: No skipping if we failed to parse the location.
+        let scopeRef = loc == nil || skip(",") ? try parseScopeRef() : nil
+        // We've skipped the comma, so failing to parse any of those two
+        // components is an error.
+        guard scopeRef != nil || loc != nil else {
+            throw parseError("Failed to parse debug location")
```
burmako: "Failed to parse source info"?

Sources/SIL/SILParser.swift
```diff
@@ -308,18 +308,20 @@ class SILParser: Parser {
...
-            let declRef = try parseDeclRef()
-            try take(":")
-            let identifier = try parseIdentifier()
-            return .case(declRef, identifier)
-        } else if skip("default") {
-            let identifier = try parseIdentifier()
-            return .default(identifier)
-        } else {
-            throw parseError("unknown case")
+    func parseCase() throws -> Case? {
```
burmako: Let's have a test for that as well?

Sources/SIL/Parser.swift
```diff
@@ -48,6 +48,17 @@ class Parser {
         return chars.suffix(from: cursor).starts(with: Array(query))
     }
 
+    // Applies the function, but restores the cursor from before the call if it returns nil.
+    func tryParse<T>(_ f: () throws -> T?) rethrows -> T? {
```
burmako: Most of the parse routines that we have are `tryParse` in the sense that they throw. How about renaming to `maybeParse`?

burmako: Also I think it would be useful to move this one next to other `parseFoo` combinators.

Sources/SIL/Parser.swift
```diff
@@ -141,6 +152,14 @@ class Parser {
         return try parseMany(pre, parseOne)
     }
 
+    func parseUntilNil<T>(_ parseOne: () throws -> T?) rethrows -> [T] {
```
burmako: Nice one!

Sources/SIL/SILParser.swift
```diff
@@ -580,18 +583,24 @@ class SILParser: Parser {
...
-        let ref = try parseInt()
-        return "scope " + String(ref)
+        return try parseInt()
     }
 
     // https://github.com/apple/swift/blob/master/docs/SIL.rst#basic-blocks
     func parseSourceInfo() throws -> SourceInfo? {
-        let scopeRef = try parseScopeRef()
+        // NB: The SIL docs say that scope refs precede locations, but this is
+        //     not true once you look at the compiler outputs or its source code.
```
burmako: Oooffff.

Sources/SIL/SILParser.swift
apaszke-g: Oh yeah I forgot to add one

Sources/SIL/SILParser.swift
apaszke-g: I know, right?

PR #272, opened 2019-09-10
Improve recovery from instruction body failures

apaszke-g:


apaszke-g:
Kokoro doesn't seem to trigger, but I've ran the tests locally and they seem to be passing.

PR #277, opened 2019-09-16
Adding xcode release v0.5-rc0 to  Installation.md

bgogul:


sinamoeini:
I tried this osx package just now and unfortunately lldb python package is configured with python 2.7. Which is a shame because jupyter requires it to be importable to python 3.

sinamoeini:
Also importing lldb in python2.7 gives the following error 
```
Sinas-MBP:~ sina$ export PYTHONPATH=`lldb -P`; python -c "import lldb"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Library/Developer/Toolchains/swift-tensorflow-RELEASE-0.5.xctoolchain/System/Library/PrivateFrameworks/LLDB.framework/Resources/Python/lldb/__init__.py", line 1510, in <module>
    class SBAddress(object):
  File "/Library/Developer/Toolchains/swift-tensorflow-RELEASE-0.5.xctoolchain/System/Library/PrivateFrameworks/LLDB.framework/Resources/Python/lldb/__init__.py", line 1681, in SBAddress
    __swig_getmethods__["module"] = GetModule
NameError: name '__swig_getmethods__' is not defined
```

dan-zheng:
> I tried this osx package just now and unfortunately lldb python package is configured with python 2.7. Which is a shame because jupyter requires it to be importable to python 3.

This is unfortunately true.

For a time, macOS development toolchains were built with Python 3. But this was not robust because Python 3 was an ad-hoc dependency (installed via Homebrew on toolchain-building machines) and caused issues for macOS users that didn't install the exact version of Python 3.

Now, macOS development toolchains are built with Python 2, which comes installed with macOS and should work for all macOS users. This means `swift-jupyter` won't work with macOS toolchains, but you can try using the [Docker installation](https://github.com/google/swift-jupyter#option-3-using-the-docker-container) instead.

sinamoeini:
Actually the last development toolchain (August 28, 2019) was built for python3 and for me it worked very nice with jupyter. I just had to relink _lldb.so to my python3 installation.

dan-zheng:
> Also importing lldb in python2.7 gives the following error
> 
> ```
> Sinas-MBP:~ sina$ export PYTHONPATH=`lldb -P`; python -c "import lldb"
> Traceback (most recent call last):
>   File "<string>", line 1, in <module>
>   File "/Library/Developer/Toolchains/swift-tensorflow-RELEASE-0.5.xctoolchain/System/Library/PrivateFrameworks/LLDB.framework/Resources/Python/lldb/__init__.py", line 1510, in <module>
>     class SBAddress(object):
>   File "/Library/Developer/Toolchains/swift-tensorflow-RELEASE-0.5.xctoolchain/System/Library/PrivateFrameworks/LLDB.framework/Resources/Python/lldb/__init__.py", line 1681, in SBAddress
>     __swig_getmethods__["module"] = GetModule
> NameError: name '__swig_getmethods__' is not defined
> ```

This is an orthogonal error, likely due to Swig version 4.0.0 on the toolchain-building machine. Swig 4.0.0 is known to cause this issue.

```console
$ head -n2 /Library/Developer/Toolchains/swift-tensorflow-RELEASE-0.5.xctoolchain/System/Library/PrivateFrameworks/LLDB.framework/Resources/Python/lldb/__init__.py
# This file was automatically generated by SWIG (http://www.swig.org).
# Version 4.0.0
```

I believe downgrading to Swig 3 may resolve the issue ([reference](https://github.com/rust-lang/rust/pull/61827/files)).
We will verify this and fix in a new v0.5 RC toolchain.

ematejska:
Since the default installed version for MacOS is 2.7 right now, it was a temporary mistake to link in 3.x into the dev versions.  In the future, we might move to supporting 3 once we can communicate to everyone to install that version for MacOS development.

sinamoeini:
Thanks for the explanation.

I tried to build the whole swift for TensorFlow toolchain myself. It was built successfully. But the completion functionality did not work. The reason was that lldb API did not have CompleteCode attribute see ([TF-743](https://bugs.swift.org/projects/TF/issues/TF-743?filter=allissues)). I looked into the swift-lldb repository. It is added to swig interface file [SBTarget.i](https://github.com/apple/swift-lldb/blob/tensorflow/scripts/interface/SBTarget.i) but it is not exposed in [LLDBWrapPython.cpp](https://github.com/apple/swift-lldb/blob/tensorflow-merge/scripts/Python/static-binding/LLDBWrapPython.cpp). Is there a way to fix this?

ematejska:
Which branch did you build the toolchain from? The tensorflow-0.5 branch should have fixes for this.  @marcrasi Do you know more?

ematejska:
Actually, https://github.com/apple/swift-lldb/pull/1961 should be in the tensorflow branch as well.

sinamoeini:
I tried tensorflow branch yesterday. @marcrasi did do the correction to swig interface file that I mentioned above

marcrasi:
I think I understand the issue with `scripts/Python/static-binding/LLDBWrapPython.cpp`.

This file is supposed to be generated by SWIG when you do a build. But it's also checked in, so that the build can run on machines without SWIG installed (see [description here](https://github.com/apple/swift-lldb/blob/3a3b6fac3df4ea365bb9d5dc9100a9edbbadf73c/scripts/prepare_bindings.py#L161)).

Since we have always been building on machines with SWIG installed, we never noticed that the checked-in generated file had gotten stale.

If my understanding is correct, you should be able to fix your build by installing SWIG before building. (I recommend version 3.0.12 because 4.x seems to cause some problems).

I'll also send a PR updating the checked-in generated files soon.

sinamoeini:
Excellent! Thank you very much @marcrasi.

marcrasi:
PR that regenerates the files: https://github.com/apple/swift-lldb/pull/1989

PR #278, opened 2019-09-16
Separate terminators from regular instructions

apaszke-g:


Sources/SIL/Printer.swift
```diff
@@ -50,6 +50,12 @@ class Printer: CustomStringConvertible {
         print(suf)
     }
 
+    func print<S: Collection>(_ xs: S, _ fn: (S.Element) -> Void) {
```
apaszke-g: I've added this because the current behavior of `print` is weird: `print(collection, "") { print($0) }` is equivalent to this overload, while `print(collection) { print($0) }` will print `collection.description`, which includes e.g. `[` and `]` for arrays.

Sources/SIL/SILParser.swift
```diff
@@ -1,4 +1,21 @@
...
+      case instruction(InstructionDef)
+      case terminator(TerminatorDef)
+
+      var description: String {
+        switch self {
+        case let .instruction(instructionDef): return instructionDef.description
+        case let .terminator(terminatorDef): return terminatorDef.description
+        }
+      }
+    }
```
burmako: How do you feel about the following data model:

```
// SIL.swift
public enum InstructionDef { // formerly, InstructionOrTerminatorDef
  case nonTerminator(NonTerminatorDef)
  case terminator(TerminatorDef)
}

public enum NonTerminatorDef { // formerly, InstructionDef
  ...
}

public enum TerminatorDef {
  ...
}

public enum Instruction { // formerly, InstructionOrTerminator
  case nonTerminator(NonTerminator)
  case terminator(Terminator)
}

public enum NonTerminator { // formerly, Instruction
  ...
}

public enum Terminator {
  ...
}
```

Some aspects of the design that may be beneficial:
  1) Hopefully clearer naming (terminators are instructions as well).
  2) All data structures live in SIL.swift so that parsing and prettyprinting work the same way for them as for other data structures of the data model (both InstructionDef and Instruction will become prettyprintable via Printer, e.g. like TerminatorDef is right now).

Sources/SIL/SIL.swift
```diff
@@ -279,16 +269,45 @@ public enum Instruction {
...
-    case unreachable
-
     // https://github.com/apple/swift/blob/master/docs/SIL.rst#witness-method
     // witness_method $Self, #Comparable."<="!1 : <Self where Self : Comparable> (Self.Type) -> (Self, Self) -> Bool : $@convention(witness_method: Comparable) <τ_0_0 where τ_0_0 : Comparable> (@in_guaranteed τ_0_0, @in_guaranteed τ_0_0, @thick τ_0_0.Type) -> Bool
     // TODO(#28): Figure out witness_method.
     case witnessMethod(_ archeType: Type, _ declRef: DeclRef, _ declType: Type, _ type: Type)
 }
 
+// https://github.com/apple/swift/blob/master/docs/SIL.rst#terminators
+public enum Terminator {
```
burmako: Is `Terminator` currently `CustomStringConvertible`? I think it should be.

Sources/SIL/SILPrinter.swift
```diff
@@ -17,9 +17,11 @@ class SILPrinter: Printer {
     func print(_ block: Block) {
         print(block.identifier)
         print(whenEmpty: false, "(", block.arguments, ", ", ")") { print($0) }
-        print(":\n")
+        print(":")
         indent()
-        print(block.instructionDefs, "\n") { print($0) }
+        print(block.instructionDefs) { print("\n"); print($0) }
+        print("\n")
+        print(block.terminatorDef)
```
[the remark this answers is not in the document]
apaszke-g: Weirdness is certainly not a blocker, but I've found this to be a more readable solution where the state both before and after instruction defs are printed is that there is no newline.

Tests/SILTests/SILParserTests.swift
```diff
@@ -18,7 +20,9 @@ public final class SILParserTests: XCTestCase {
         let instr = "%122 = apply garbage..."
         let parser = SILParser(forString: instr)
         do {
-            let def = try parser.parseInstructionDef()
+            guard case let .instruction(def) = try parser.parseInstructionDef() else {
+                return XCTFail("Expected the result to be an instruction")
```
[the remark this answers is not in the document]
apaszke-g: Sure, I'll change it to `non-terminating instruction`.

Sources/SIL/SILParser.swift
apaszke-g: I feel like this will add a lot of verbosity. The only place where the instruction/terminator confusion arises is inside the parser when you never know what will come next. Your proposal propagates the same confusion throughout the whole system. For example, it does not guarantee that all defs are non-terminators and that a block ends with a terminator. Additionally, all of the pattern matching code (e.g. for abstract interpretation) now becomes a level deeper in user code, simply because it has to handle the incorrect "is a terminator" case, which should not appear in the body of the basic block anyway.

Given the added verbosity and weaker type safety I would rather keep what's there right now.

Sources/SIL/SIL.swift
apaszke-g: Good point, will make it so.

Sources/SIL/SILParser.swift
burmako: I was in a rush when writing this comment, so I didn't explain the idea well enough. We had a brief internal chat, and the future version of this PR will reflect a similar design.

PR #281, opened 2019-09-19
Add support for value name substitutions

apaszke-g:
This is useful when one wants to actually edit the loaded SIL by e.g. inlining a function into the other one, unroll a loop, or simply duplicate a block.

### DIFF
--- a/Sources/SIL/SIL.swift
+++ b/Sources/SIL/SIL.swift
@@ -1,5 +1,5 @@
 // https://github.com/apple/swift/blob/master/docs/SIL.rst
-public class Module {
+public final class Module {
     public var functions: [Function]
 
     public init(_ functions: [Function]) {
@@ -18,7 +18,7 @@ public class Module {
 }
 
 // https://github.com/apple/swift/blob/master/docs/SIL.rst#functions
-public class Function {
+public final class Function {
     public var linkage: Linkage
     public var attributes: [FunctionAttribute]
     public var name: String
@@ -38,7 +38,7 @@ public class Function {
 }
 
 // https://github.com/apple/swift/blob/master/docs/SIL.rst#basic-blocks
-public class Block: Equatable {
+public final class Block: Equatable {
     public var identifier: String
     public var arguments: [Argument]
     public var operatorDefs: [OperatorDef]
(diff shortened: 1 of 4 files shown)


Sources/SIL/SIL.swift
```diff
@@ -38,7 +38,7 @@ public class Function {
 }
 
 // https://github.com/apple/swift/blob/master/docs/SIL.rst#basic-blocks
-public class Block: Equatable {
+public final class Block: Equatable {
```
apaszke-g: This needs to be `final` or otherwise I won't be able to conform `Block` to `CanSubstituteValueNames`, because then Swift complains that there's no guarantee that `Self` is actually a `Block` (which it really is, but I guess it would like the type to always get refined to a potential subclass...).

Sources/SIL/SILAnalysis.swift
```diff
@@ -1,57 +1,186 @@
...
-        case let .store(value, _, operand): return [value, operand.value]
-        case .stringLiteral(_, _): return []
-        case let .strongRelease(operand): return [operand.value]
-        case let .strongRetain(operand): return [operand.value]
-        case let .struct(_, operands): return unwrap(operands)
-        case let .structElementAddr(operand, _): return [operand.value]
-        case let .structExtract(operand, _): return [operand.value]
-        case let .thinToThickFunction(operand, _): return [operand.value]
+        case .allocStack(_, _): return self
+        case let .apply(nothrow, value, substitutions, arguments, type):
```
burmako: Can types refer to names that are being substituted?

Sources/SIL/SILAnalysis.swift
```diff
@@ -1,57 +1,186 @@
...
+            return .apply(nothrow, s(value), substitutions, arguments.map(s), type)
+        case .beginAccess(_, _, _, _, _): return self
```
burmako: Looks like `beginAccess.operand` is unhandled here.

Sources/SIL/SILAnalysis.swift
apaszke-g: No, as otherwise you'd have dependent types.

Sources/SIL/SILAnalysis.swift
apaszke-g: Nice catch! Added a test that should ensure we won't have problems like this in the future, and fixed the error!

PR #283, opened 2019-09-19
Implement Operation.operands and Terminator.operands

apaszke-g:
Apparently some transformations are best done when you have the type information available.

I'm not really sure what's the best way to test this, so I guess I'll just add a few checks at least for non-trivial cases in a subsequent commit. This is meant to be slightly WIP, but I'm putting it out for comments.

### DIFF
--- a/Sources/SIL/SIL.swift
+++ b/Sources/SIL/SIL.swift
@@ -495,6 +495,7 @@ public enum TupleElements: Equatable {
 public indirect enum Type: Equatable {
     case addressType(_ type: Type)
     case attributedType(_ attributes: [TypeAttribute], _ type: Type)
+    case coroutineTokenType
     case functionType(_ parameters: [Type], _ result: Type)
     case genericType(_ parameters: [String], _ requirements: [TypeRequirement], _ type: Type)
     case namedType(_ name: String)
--- a/Sources/SIL/SIL.swift
+++ b/Sources/SIL/SIL.swift
@@ -84,6 +84,13 @@ public struct TerminatorDef: Equatable {
 public enum InstructionDef: Equatable {
     case `operator`(OperatorDef)
     case terminator(TerminatorDef)
+
+    var instruction: Instruction {
+        switch self {
+        case let .operator(def): return .operator(def.operator)
+        case let .terminator(def): return .terminator(def.terminator)
+        }
+    }
 }
 
 
--- a/Sources/SIL/SILAnalysis.swift
+++ b/Sources/SIL/SILAnalysis.swift
@@ -228,56 +228,56 @@ extension Type {
             return .withOwnership(attribute, subtype.substituted(using: s))
         }
     }
-}
 
-func specialize(_ type: Type, to arguments: [Type]) -> Type {
-    switch type {
-    case let .addressType(subtype):
-        return .addressType(specialize(subtype, to: arguments))
-    case let .attributedType(attributes, subtype):
-        return .attributedType(attributes, specialize(subtype, to: arguments))
-    case let .genericType(parameters, _, subtype):
-        guard parameters.count == arguments.count else {
-            fatalError(
-                "Specializing a generic type with \(parameters.count) parameters using \(arguments.count) arguments"
-            )
+    func specialized(to arguments: [Type]) -> Type {
+        switch self {
+        case let .addressType(subtype):
+            return .addressType(subtype.specialized(to: arguments))
+        case let .attributedType(attributes, subtype):
+            return .attributedType(attributes, subtype.specialized(to: arguments))
+        case let .genericType(parameters, _, subtype):
+            guard parameters.count == arguments.count else {
+                fatalError(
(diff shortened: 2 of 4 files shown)


Sources/SIL/SILAnalysis.swift
```diff
@@ -46,7 +46,8 @@ extension Operator: AlphaConvertible {
         case let .apply(nothrow, value, substitutions, arguments, type):
             return .apply(nothrow, s(value), substitutions, arguments.map(s), type)
         case let .beginAccess(access, enforcement, noNestedConflict, builtin, operand):
-            return .beginAccess(access, enforcement, noNestedConflict, builtin, operand.alphaConverted(using: s))
+            return .beginAccess(
```
apaszke-g: This is `swift-format`, not a meaningful change

Sources/SIL/SIL.swift
```diff
@@ -495,6 +495,7 @@ public enum TupleElements: Equatable {
 public indirect enum Type: Equatable {
     case addressType(_ type: Type)
     case attributedType(_ attributes: [TypeAttribute], _ type: Type)
+    case coroutineTokenType
```
apaszke-g: I had to create this, because the coroutine tokens never appear annotated with a type (they can only be results of `begin_apply` and sole arguments to `end_apply`). However, it's convenient to be able to wrap them into `Operand`s, so this is a placeholder type for them.

Sources/SIL/SILAnalysis.swift
```diff
@@ -195,11 +196,193 @@ extension Array: AlphaConvertible where Element: AlphaConvertible {
...
+        return destructFunctionType(subtype)
+    case let .withOwnership(_, subtype):
+        return destructFunctionType(subtype)
+    case .addressType(_): fallthrough
+    case .coroutineTokenType: fallthrough
+    case .namedType(_): fallthrough
+    case .selectType(_, _): fallthrough
+    case .selfType: fallthrough
+    case .specializedType(_, _): fallthrough
+    case .tupleType(_):
```
burmako: I think you can do something like `case .foo(_), .bar(_, _): ...`

Sources/SIL/SILAnalysis.swift
```diff
@@ -195,11 +196,193 @@ extension Array: AlphaConvertible where Element: AlphaConvertible {
     }
 }
 
+func substitute(_ type: Type, using s: (String) -> Type) -> Type {
```
burmako: Let's have this as an extension method on `Type` to be consistent with things like `Operator.alphaConverted` etc?

Sources/SIL/SILAnalysis.swift
```diff
@@ -195,11 +196,193 @@ extension Array: AlphaConvertible where Element: AlphaConvertible {
...
+        return .specializedType(
+            substitute(genericType, using: s), arguments.map { substitute($0, using: s) })
+    case let .tupleType(elementTypes):
+        return .tupleType(elementTypes.map { substitute($0, using: s) })
+    case let .withOwnership(attribute, subtype):
+        return .withOwnership(attribute, substitute(subtype, using: s))
+    }
+}
+
+func specialize(_ type: Type, to arguments: [Type]) -> Type {
```
burmako: Ditto.

Sources/SIL/SILAnalysis.swift
burmako: Perhaps, it's the time to have more strongly-typed names? Currently, we use String for a lot of things (e.g. both local and global names), so maybe it would be useful to make flavors of names apparent in the type system?

Sources/SIL/SILPrinter.swift
```diff
@@ -537,6 +537,8 @@ class SILPrinter: Printer {
         case let .attributedType(attrs, type):
             print("", attrs, " ", " ") { print($0) }
             naked(type)
+        case .coroutineTokenType:
+            print("!CoroutineTokenType!")
```
burmako: This won't parse, right?

Sources/SIL/SILAnalysis.swift
apaszke-g: I remember trying this once, and it didn't work, so my conclusion was that this chaining is limited to e.g. integer literals, but I might be wrong.

Sources/SIL/SILAnalysis.swift
apaszke-g: Sure, can do.

apaszke-g: Please, yes. I already have a few `newtype`-like wrappers in my project, and at least having type aliases would be helpful. Having said this, it's out of scope of this commit.

Sources/SIL/SILPrinter.swift
apaszke-g: It won't, but it also doesn't have to. There's no way to express this type in the textual SIL format.

Sources/SIL/SILAnalysis.swift
burmako: Doesn't have to be done in this PR, I agree.

PR #297, opened 2019-10-04
Update testFor for changes in the AST shape upstream

burmako:


burmako:
Kokoro is failing because it uses a development S4TF toolchain that uses an older compiler that is still producing an old AST shape. Let's wait until a new development S4TF toolchain gets built and propagated to Kokoro.

burmako:
There seems to be a circular dependency between the libQuote presubmit and the development toolchain job. Therefore, I'm going to merge this PR with the presubmit broken to resolve the stalemate.

PR #300, opened 2019-10-11
Skeleton of the .swiftmodule parser

apaszke:
This is a very WIP version of the .swiftmodule parser that I've implemented on top of the LLVM bitcode parser. IIRC it's mostly done, except that it cannot parse the types. I'm putting it out here so that the code doesn't get lost, but I'm not sure if we really want to land this in this state. I'm leaving the final decision up to @burmako.

### DIFF
--- a/Sources/SIL/SIBParser.swift
+++ b/Sources/SIL/SIBParser.swift
@@ -1,0 +1,399 @@
+import Foundation
+
+let SIB_MAGIC: [UInt8] = [0xE2, 0x9C, 0xA8, 0x0E]
+
+enum SIBFileError: Error {
+  case cannotOpenFile
+  case incorrectMagic
+}
+
+enum SIBError: Error {
+  case parseError(_ reason: String)
+  case unsupported(_ what: String)
+}
+
+func loadSIBBitcode(fromPath path: String) throws -> BitcodeBlock {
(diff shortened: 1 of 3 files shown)


Sources/SIL/SIBParser.swift
```diff
@@ -0,0 +1,399 @@
...
+  var stream = Bitstream(handle.readDataToEndOfFile())
+  if (try stream.next(bytes: 4) != SIB_MAGIC) {
+    throw SIBFileError.incorrectMagic
+  }
+
+  let parser = BitcodeParser(stream)
+  return try parser.parse()
+}
+
+fileprivate func expectOne<T>(called name: String, among candidates: [T]) throws -> T {
```
rxwei:
```suggestion
fileprivate func expectOne<T>(named name: String, among candidates: [T]) throws -> T {
```
It's usually "named:" rather than "called:", e.g. [`Image.init(named:)`](https://developer.apple.com/documentation/uikit/uiimage/1624146-init) :)

burmako:
It would be good to finish up this PR (implement support for parsing types, like Adam says, then write  a bit more tests), but it looks like we didn't have time to dedicate to that in a while. I'm going to close this PR, and we can revive it when the time comes.

PR #305, opened 2019-10-15
Use `python3 -m pip` instead of pip3

yashk2810:
There is some weird thing going on with pip3 when its upgraded. This gets around that.


yashk2810:
@marcrasi Can you take a look please?

PR #311, opened 2019-10-27
Update testThrow in StructureTests in libQuote

burmako:

PR #315, opened 2019-11-06
temporarily disable testThrow for merge

marcrasi:
The merge from upstream causes this test to fail, because the second thing inside `Name` has changed from `""` to `"<unstable USR>"`.

Assuming this isn't a regression, the right thing to do is to update the assertion. But if I update the assertion now, our toolchain builds will start failing. So I'd like to disable this test, finish the merge, then reenable this test with the new assertion.

Question: Does this mean that TF-937 is fixed after the merge? It looks like this is what TF-937 is talking about, but I'm not sure.

```
Test Case 'StructureTests.testThrow' started at 2019-11-05 20:35:47.228
/swift/Tests/QuoteTests/StructureTests.swift:449: error: StructureTests.testThrow : XCTAssertEqual failed: ("Closure(
  [],
  [Throw(
    Conversion(
      Call(
        Name(
          "X",
          "<unstable USR>",
          FunctionType(
            [],
            [],
            TypeName("X", "<unstable USR>"))),
        [],
        [],
        TypeName("X", "<unstable USR>")),
      TypeName("Error", "s:s5ErrorP")))],
  FunctionType(
    [],
    [],
    TupleType(
      [])))") is not equal to ("Closure(
  [],
  [Throw(
    Conversion(
      Call(
        Name(
          "X",
          "",
          FunctionType(
            [],
            [],
            TypeName("X", "<unstable USR>"))),
        [],
        [],
        TypeName("X", "<unstable USR>")),
      TypeName("Error", "s:s5ErrorP")))],
  FunctionType(
    [],
    [],
    TupleType(
      [])))") -
Test Case 'StructureTests.testThrow' failed (0.0 seconds)
```

### DIFF
--- a/Tests/QuoteTests/StructureTests.swift
+++ b/Tests/QuoteTests/StructureTests.swift
@@ -441,38 +441,39 @@ public final class StructureTests: XCTestCase {
         )
     }
 
-    public func testThrow() {
-        let q = #quote{
-            throw X()
-        }
-        // TODO(TF-937): Fix the empty symbol in the Name's structure below.
-        assertStructure(
-            q,
-            """
-      Closure(
-        [],
-        [Throw(
-          Conversion(
-            Call(
-              Name(
-                "X",
-                "",
-                FunctionType(
-                  [],
-                  [],
-                  TypeName("X", "<unstable USR>"))),
-              [],
-              [],
-              TypeName("X", "<unstable USR>")),
-            TypeName("Error", "s:s5ErrorP")))],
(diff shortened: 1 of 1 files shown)


burmako:
Looks like there was a temporary change in ASTs upstream, which first prompted us to update `"<unstable USR>"` to `""` (https://github.com/tensorflow/swift/pull/311) and now prompted us to revert the update. If the warning is gone, then TF-937 should be fixed.

PR #317, opened 2019-11-07
Release 2019-11-07 toolchain for macOS.

dan-zheng:
https://github.com/apple/swift/commit/4d1c1df58f9aeb0bc2bdba0b6723d1646fb0ddab

### DIFF
--- a/Installation.md
+++ b/Installation.md
@@ -54,7 +54,7 @@ To install Swift for TensorFlow, download one of the packages below and follow t
 
 | Download | Date |
 |----------|------|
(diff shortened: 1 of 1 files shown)


dan-zheng:
Please hold off on merging, the toolchain may have incorrectly been built with Python 3:
```
$ dyld: Library not loaded: /Library/Frameworks/Python.framework/Versions/3.5/Python
  Referenced from: /Library/Developer/Toolchains/swift-tensorflow-DEVELOPMENT-2019-11-07-a.xctoolchain/System/Library/PrivateFrameworks/LLDB.framework/Versions/A/LLDB
  Reason: image not found
[1]    63538 abort      swift
```

It's not guaranteed that users have any given version of Python 3 installed on their system, so I think building with Python 2 is the only sure way to ensure toolchains work universally.

PR #318, opened 2019-11-07
Add installation verification steps

texasmichelle:
Add a note to help first-time users who encounter REPL crashes like
https://bugs.swift.org/browse/TF-940.

### DIFF
--- a/Installation.md
+++ b/Installation.md
@@ -191,3 +191,21 @@ $ export PATH=$(pwd)/usr/bin:"${PATH}"
 You can now execute the `swift` command to run the REPL or build Swift projects.
 
 **Note:** If you are using a CUDA build and you have an NVIDIA GPU with a compute capability other than 3.5 or 7.0, then you will experience a ~10 minute delay the first time you execute a TensorFlow operation, while TensorFlow compiles kernels for your GPU's compute capability. The program will not print anything out and it will appear to be frozen.
+
+# Verify the Installation
+
+Create a text file `test.swift` with the following contents:
+
+```
+import TensorFlow
+var x = Tensor<Float>([[1, 2], [3, 4]])
+print(x + x)
+```
+
+Run this command to verify the installation. If you see this output, you have successfully installed Swift for TensorFlow!
+
+```
+$ swift test.swift
+[[2.0, 4.0],
+ [6.0, 8.0]]
+```
--- a/Installation.md
+++ b/Installation.md
@@ -132,12 +132,12 @@ To install Swift for TensorFlow, download one of the packages below and follow t
 
 6. Selecting a Swift toolchain affects the Xcode IDE only. To use the Swift toolchain with command-line tools, add the Swift toolchain to your path as follows:
 
-    ```bash
+    ```console
     $ export PATH=/Library/Developer/Toolchains/swift-latest/usr/bin:"${PATH}"
     ```
 
 7. **CUDA-only**: If you downloaded a CUDA GPU version of the toolchain, add the library path(s) to CUDA and CuDNN to `$LD_LIBRARY_PATH`.
-    ```bash
+    ```console
     $ export LD_LIBRARY_PATH=/usr/local/cuda/lib:"${LD_LIBRARY_PATH}"
     ```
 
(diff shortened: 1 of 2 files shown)


Usage.md
```diff
@@ -50,6 +50,10 @@ $R1: TensorFlow.Tensor<Float> = [[8.0, 16.0], [24.0, 32.0]]
 $R2: TensorFlow.Tensor<Float> = [32.0, 48.0]
 ```
 
+**Note:** If you encounter issues with the REPL crashing,
```
pschuh: Perhaps it is better to just reference https://bugs.swift.org/browse/TF-940 as a known issue so that when it gets fixed it is clear that this section can be deleted?

Installation.md
```diff
@@ -191,3 +191,21 @@ $ export PATH=$(pwd)/usr/bin:"${PATH}"
 You can now execute the `swift` command to run the REPL or build Swift projects.
 
 **Note:** If you are using a CUDA build and you have an NVIDIA GPU with a compute capability other than 3.5 or 7.0, then you will experience a ~10 minute delay the first time you execute a TensorFlow operation, while TensorFlow compiles kernels for your GPU's compute capability. The program will not print anything out and it will appear to be frozen.
+
+# Verify the Installation
+
+Create a text file `test.swift` with the following contents:
+
+```
```
dan-zheng: Please use Swift syntax highlighting:
```suggestion
```swift
```

Installation.md
```diff
@@ -191,3 +191,21 @@ $ export PATH=$(pwd)/usr/bin:"${PATH}"
...
+import TensorFlow
+var x = Tensor<Float>([[1, 2], [3, 4]])
+print(x + x)
+```
+
+Run this command to verify the installation. If you see this output, you have successfully installed Swift for TensorFlow!
+
+```
```
dan-zheng: Please use `console` syntax highlighting for code snippets from the console, for consistency with the rest of the document.

Usage.md
```diff
@@ -50,6 +50,10 @@ $R1: TensorFlow.Tensor<Float> = [[8.0, 16.0], [24.0, 32.0]]
...
+[verify your installation](./Installation.md#verify-the-installation) and
```
dan-zheng: Minor: consistency naming

```suggestion
[verify the installation](./Installation.md#verify-the-installation) and
```

Usage.md
dan-zheng: I think that's a good idea! How about this:
```suggestion
    **Note:** using the `TensorFlow` module in the Swift REPL on macOS is known to
    be problematic since Swift for TensorFlow 0.5.

    ```console
    $ swift
    Welcome to Swift version 5.1-dev (LLVM 200186e28b, Swift 1238976565).
    Type :help for assistance.
     1> import TensorFlow
     2> Tensor(1)
    error: Couldn't lookup symbols:
     TensorFlow.TensorHandle.init(copyingFromCTensor: Swift.OpaquePointer) -> TensorFlow.TensorHandle<A>
     TensorFlow.TensorHandle.init(copyingFromCTensor: Swift.OpaquePointer) -> TensorFlow.TensorHandle<A>
    ```

    [TF-940](https://bugs.swift.org/browse/TF-940) tracks this issue.
```

texasmichelle: My intention with this phrasing was to channel users to the bug for upvoting and questions/comments. Even after this bug is resolved, it's not unreasonable to expect a different bug to crop up that manifests in a similar way. Even if we forget to remove this from the docs, it could be useful as an example of an issue the user could generate.

That being said, it looks like I included the wrong link to the project itself rather than the specific bug. I'm fine with changing the phrasing so that it's more obvious when it can be removed.

Installation.md
texasmichelle: There is inconsistency among both files, so I'll update everything to use `console` instead of `bash` or empty.

Usage.md
texasmichelle: Your phrasing is good @dan-zheng - I just moved the issue link closer to the Note for more clarity.

texasmichelle:
Appreciate the swift review!! (Couldn't help myself considering it's my first swift PR 😂)

PR #330, opened 2019-11-25
Update DifferentialTypes docs

steph-en-m:
The proposed change acknowledges that not all functions have derivatives as currently outlined in the preface of this doc.

### DIFF
--- a/docs/DifferentiableTypes.md
+++ b/docs/DifferentiableTypes.md
@@ -15,7 +15,7 @@ Last updated: March 2019
 
 ## Preface
 
-Speaking in terms of elementary calculus, only functions are "differentiable": only functions have derivatives and can be differentiated. In this document, the terminology "differentiable types" is used as a shorthand for "types that can be used as arguments and results of differentiable functions". This notion is important because not all types are "differentiable" in this sense. For example, types representing real numbers and vector spaces are "differentiable", but strings and integers are not.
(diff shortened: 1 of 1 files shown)


docs/DifferentiableTypes.md
```diff
@@ -15,7 +15,7 @@ Last updated: March 2019
 
 ## Preface
 
-Speaking in terms of elementary calculus, only functions are "differentiable": only functions have derivatives and can be differentiated. In this document, the terminology "differentiable types" is used as a shorthand for "types that can be used as arguments and results of differentiable functions". This notion is important because not all types are "differentiable" in this sense. For example, types representing real numbers and vector spaces are "differentiable", but strings and integers are not.
+Speaking in terms of elementary calculus, only functions are "differentiable": only functions that have derivatives can be differentiated. In this document, the terminology "differentiable types" is used as a shorthand for "types that can be used as arguments and results of differentiable functions". This notion is important because not all types are "differentiable" in this sense. For example, types representing real numbers and vector spaces are "differentiable", but strings and integers are not.
```
dan-zheng: I think "only functions can *have derivatives*" is closer to the original intention:
```suggestion
Speaking in terms of elementary calculus, only functions are "differentiable": only functions can *have derivatives* and can *be differentiated*. In this document, the terminology "differentiable types" is used as a shorthand for "types that can be used as arguments and results of differentiable functions". This notion is important because not all types are "differentiable" in this sense. For example, types representing real numbers and vector spaces are "differentiable", but strings and integers are not.
```

steph-en-m: Alright.

PR #352, opened 2019-12-30
Adds a step for bypassing code signing problem

rahulbhalley:
Users are experiencing problem with executing S4TF program in their Xcode project as being discussed in issue #177. 

This PR adds a new installation step that resolves the issue #177 by enabling the **Disable Library Validation** option from **Signing & Capabilities > Hardened Runtime**. This option lets the Xcode project to add frameworks without requiring any code signing. Please see the official [documentation](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_disable-library-validation) for details.

cc @dan-zheng @rxwei

PR #362, opened 2020-01-11
Remove obsolete Xcode 10 usage instructions.

dan-zheng:
Using Xcode's legacy build system is no longer required.


dan-zheng:
Happy to address suggestions later! Merging to unblock other documentation improvements (https://github.com/tensorflow/swift/pull/352).

Issue #350, opened 2019-12-24
macOS toolchains do not work with Xcode new build system


dan-zheng:
Various errors occur using Swift for TensorFlow macOS toolchains with Xcode's new build system:

![image](https://user-images.githubusercontent.com/5590046/71424898-da57e580-2664-11ea-9038-80c04a101fb3.png)

dan-zheng:
The cause is that the `Version` string in `Info.plist` provided in toolchains cannot be parsed by Xcode:
```
$ /Applications/Xcode.app/Contents/MacOS/Xcode Foo.xcodeproj
2019-12-24 15:47:16.618 XCBBuildService[75398:12495542] /Library/Developer/Toolchains/swift-tensorflow-DEVELOPMENT-2019-12-12-a.xctoolchain: warning: failed to load toolchain: 'Version' parse error: Could not parse version component from: 'swift-tensorflow-DEVELOPMENT-2019-12-12-a
```

This is an issue in the open source [`utils/build-toolchain`](https://github.com/apple/swift/blob/746b58e8e1c1ef1ac09d8031875fd2a08b65597c/utils/build-toolchain#L129) script. Someone discovered this issue [on the forums](https://forums.swift.org/t/xcode-wont-pick-up-a-custom-swift-toolchain/12825) but didn't submit a fix.

dan-zheng:
Fixed in https://github.com/apple/swift/pull/28945.

New macOS toolchain released: https://storage.googleapis.com/swift-tensorflow/mac/swift-tensorflow-DEVELOPMENT-2019-12-23-a-osx.pkg

![image](https://user-images.githubusercontent.com/5590046/71425018-38d19380-2666-11ea-8246-7ded8dfe6958.png)

PR #372, opened 2020-01-28
Mention the legacy build system option

BrunoBerisso:
I just install the toolchain (Xcode 11.3.1 - TF 0.6.0) and got _No such module TensofrFlow_ error. Looking through the issues there are 3 or 4 related to this error all fixed by setting the _Build System_ to `Legacy Build System` instead of the `New Build System` which is the default now.

This fix also works for me so I thought it could be a good idea to have it mentioned in the instructions.

Thanks for this repo!

### DIFF
--- a/Installation.md
+++ b/Installation.md
@@ -163,6 +163,8 @@ To install Swift for TensorFlow, download one of the packages below and follow t
     $ export LD_LIBRARY_PATH=/usr/local/cuda/lib:"${LD_LIBRARY_PATH}"
     ```
 
+**Note:** if after selecting the TensorFlow toolchain in Xcode you still get "No such module 'TensorFlow'" error try setting the legacy build system under `File` -> `Project Settings...` -> `Build System`
+
 ## Linux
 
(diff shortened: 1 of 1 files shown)


8bitmp3:
See an earlier PR from the dev team here https://github.com/tensorflow/swift/pull/362 (Jan 11, 2020): “Using Xcode's legacy build system is no longer required.”

Dan addressed the "macOS toolchains do not work with Xcode new build system" issue - https://github.com/tensorflow/swift/issues/350 - in late December.

If you download one of the latest dev toolchains (Dec 23 onwards) switching to Legacy shouldn’t be required.

BradLarson:
I can confirm that this shouldn't be necessary any longer for the nightly builds. We're working on a new release, and that should also alleviate the need for the legacy build system in the official release builds as well.

I'd prefer we not reintroduce these instructions, and instead focus on getting a release build out that fixes this for those toolchains as well.

BrunoBerisso:
@googlebot I signed it!

Great, thanks!

PR #375, opened 2020-01-31
Remove quasiquotes from tensorflow/swift

burmako:
Quasiquotes were introduced in tensorflow/swift as a promising approach
to metaprogramming to power our work on Swift MLIR, but we found a SIL-based
approach to be more useful for the case studies that we currently have in mind.

Aside from Swift MLIR, we don't have any other use cases for quasiquotes,
so I'm suggesting we remove them to: 1) reduce the delta between tensorflow and
master which goes along the lines of our long-term plans to converge these two
branches, 2) simplify weekly merges.


AriMKatz:
Out of curiosity, Is there any information available on the general idea behind the SIL approach and its advantages?

Issue #405, opened 2020-03-12
Python tutorial broken


ematejska:
If you run the python interoperability notebook:
https://colab.research.google.com/github/tensorflow/swift/blob/master/docs/site/tutorials/python_interoperability.ipynb

One gets the error in the first code block of:
error: <Cell 25>:2:1: error: unexpected conditional compilation block terminator
#else
^

error: <Cell 25>:4:1: error: unexpected conditional compilation block terminator
#endif
^

This was reported by a user at: https://groups.google.com/a/tensorflow.org/g/swift/c/_HMOs8UNcpo

ematejska:
If I add "import TensorFlow" before the #if-else-endif import block, it starts working again.  Why is this required?  The #if-else-endif import block was added recently as part of adding PythonKit.

ematejska:
In the very least, we should update the notebook.

marcrasi:
I was updating a different notebook this morning with exactly this problem, so I know what it is.

Colab treats the first line as a comment if it begins with "#". So adding any line above the "#if" line fixes it.

PR #406, opened 2020-03-12
fix python interoperability notebook

marcrasi:
Colab treats the first line as a comment if it begins with "#". So adding any line above the "#if" line fixes it.

Fixes https://github.com/tensorflow/swift/issues/405.


Kshitij09:
This does solve the issue. Will try doing the same while dealing with conditional blocks.

PR #426, opened 2020-04-23
simplify nightly toolchain links

marcrasi:
Now x10 is included in all the nightly Linux toolchains by default, so we can remove the separate entries for x10 toolchains.

### DIFF
--- a/Installation.md
+++ b/Installation.md
@@ -86,10 +86,8 @@ To install Swift for TensorFlow, download one of the packages below and follow t
 | Download |
 |----------|
 | [Xcode 11 (April 17, 2020)](https://storage.googleapis.com/swift-tensorflow/mac/swift-tensorflow-DEVELOPMENT-2020-04-17-a-osx.pkg) |
-| [Ubuntu 18.04 (CPU Only) (Nightly)](https://storage.googleapis.com/swift-tensorflow-artifacts/nightlies/latest/swift-tensorflow-DEVELOPMENT-ubuntu18.04.tar.gz) |
-| [Ubuntu 18.04 (x10 TPU, CPU) (Nightly)](https://storage.googleapis.com/swift-tensorflow-artifacts/nightlies/latest/swift-tensorflow-DEVELOPMENT-x10-ubuntu18.04.tar.gz) |
-| [Ubuntu 18.04 (CUDA 10.1) (Nightly)](https://storage.googleapis.com/swift-tensorflow-artifacts/nightlies/latest/swift-tensorflow-DEVELOPMENT-cuda10.1-cudnn7-ubuntu18.04.tar.gz) |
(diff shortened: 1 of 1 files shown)


asuhan:
"x10 TPU" is redundant imo, we could just say TPU.

marcrasi:
Updated to say "TPU"!

Installation.md
```diff
@@ -86,10 +86,8 @@ To install Swift for TensorFlow, download one of the packages below and follow t
 | Download |
 |----------|
 | [Xcode 11 (April 17, 2020)](https://storage.googleapis.com/swift-tensorflow/mac/swift-tensorflow-DEVELOPMENT-2020-04-17-a-osx.pkg) |
-| [Ubuntu 18.04 (CPU Only) (Nightly)](https://storage.googleapis.com/swift-tensorflow-artifacts/nightlies/latest/swift-tensorflow-DEVELOPMENT-ubuntu18.04.tar.gz) |
-| [Ubuntu 18.04 (x10 TPU, CPU) (Nightly)](https://storage.googleapis.com/swift-tensorflow-artifacts/nightlies/latest/swift-tensorflow-DEVELOPMENT-x10-ubuntu18.04.tar.gz) |
-| [Ubuntu 18.04 (CUDA 10.1) (Nightly)](https://storage.googleapis.com/swift-tensorflow-artifacts/nightlies/latest/swift-tensorflow-DEVELOPMENT-cuda10.1-cudnn7-ubuntu18.04.tar.gz) |
-| [Ubuntu 18.04 (x10 TPU, CUDA 10.1, CPU) (Nightly)](https://storage.googleapis.com/swift-tensorflow-artifacts/nightlies/latest/swift-tensorflow-DEVELOPMENT-x10-cuda10.1-cudnn7-ubuntu18.04.tar.gz) |
+| [Ubuntu 18.04 (CPU, x10 TPU) (Nightly)](https://storage.googleapis.com/swift-tensorflow-artifacts/nightlies/latest/swift-tensorflow-DEVELOPMENT-ubuntu18.04.tar.gz) |
```
ematejska: Awesome about removing the X10 builds.
I agree with Alex that x10 TPU is duplicative, so maybe just TPU?  Also, the regular builds also support GPU's, just not CUDA GPUs correct?  How can we make that clear? Is it OpenCL? I'm also ok with:

Ubuntu 18.04 (CPU, GPU, TPU)
Ubuntu 18.04 (CPU, CUDA 10.1, TPU)

What do you think?

marcrasi: Do they support computation on GPU using non-CUDA things? I have no idea.

asuhan: IIRC TensorFlow supports OpenCL through SYCL, but I don't think it's enabled by default, so we probably don't have it.

ematejska: Well then we the changes look correct given that feedback and merging.

PR #445, opened 2020-05-04
[Do Not Merge] Updating for `@derivative(of:)` usage in Raw TensorFlow Operators tutorial

BradLarson:
The  current Raw TensorFlow Operators tutorial uses the deprecated `@differentiable(vjp:)` syntax, which will be removed in the 0.9 release toolchain (and is already removed in the nightlies). This replaces that implementation and description with the modern equivalents.


ematejska:
I think we need this in the 0.9 release branch first and this outstanding.  Please don't merge until 0.9 Colab goes live.

ematejska:
Actually, does this already work in 0.8?

dan-zheng:
This patch should be valid in 0.8 and 0.9.

BradLarson:
Yes, I can confirm this works in 0.8 as well, wasn't sure if the best approach was to start here and cherry-pick back to tensorflow-0.9.

PR #449, opened 2020-05-05
Adding docs about backends

ematejska:
Now that we've added X10 and the user has choices about what backend to use, document it.


ematejska:
Brad will followup with some updates to this doc post merge.

PR #459, opened 2020-05-08
Update 0.9 macOS toolchain from rc1 to rc2.

dan-zheng:
rc2 toolchain has a fix for TF-1260: macOS 10.15+ `swiftc` binary dynamic linker issues.

I kept the macOS toolchain as a release candidate in case @ematejska wants to fix TF-1273 (immediate mode crash) before creating an official release.


dan-zheng:
Thanks for the date fix :)

PR #462, opened 2020-05-11
X10 tutorial - avoid `scalarized()` in loops for defaultXLA

mikowals:
Keeping accumulated training statistics as tensors and only calling `scalarized()` for printing appears to speed up performance ~2x with XLA devices.  The timings below were done on Colab free instances.

- seconds per epoch - CPU - XLA goes from 82 to 39 
- seconds per epoch - TPU - XLA goes from 11.8 to 6.5
- seconds per epoch - CPU - Eager - goes from 37 to 35 (probably noise, reported for comparison)

The changes I made are:

-  import Foundation to time and report seconds per epoch
-  data in Statistics struct as Tensors so `scalarized()` called outside batch loops
-  add `Statistics.init(on: device)` to avoid device mismatch errors
-  DRY accumulating batch statistics with `update()` method
-  format print strings with fewer decimal places

There may be a simpler way to get the same impact in the training tutorial.  Also, I have seen [similar Statistics code](https://github.com/tensorflow/swift-apis/blob/master/Sources/x10/swift_bindings/training_loop.swift#L122) that might be a more general solution.

### DIFF
--- a/docs/site/tutorials/introducing_x10.ipynb
+++ b/docs/site/tutorials/introducing_x10.ipynb
@@ -309,7 +309,7 @@
         "colab_type": "text"
       },
       "source": [
-        "\"Now, we will implement basic progress tracking and reporting.  All intermediate statistics are kept as tensors on the same device where training is run and `scalarized()` is called only during reporting. This will be especially important later when using X10, because it avoids unnecessary materialization of lazy tensors.\""
+        "Now, we will implement basic progress tracking and reporting.  All intermediate statistics are kept as tensors on the same device where training is run and `scalarized()` is called only during reporting. This will be especially important later when using X10, because it avoids unnecessary materialization of lazy tensors."
       ]
     },
     {
--- a/docs/site/tutorials/introducing_x10.ipynb
+++ b/docs/site/tutorials/introducing_x10.ipynb
@@ -36,26 +36,14 @@
       "metadata": {
         "colab_type": "code",
         "id": "kZRlD4utdPuX",
-        "colab": {
-          "base_uri": "https://localhost:8080/",
-          "height": 34
-        },
-        "outputId": "34002f79-193b-4341-9649-a5ac7452be20"
+        "colab": {}
       },
       "source": [
         "%install '.package(url: \"https://github.com/tensorflow/swift-models\", .branch(\"tensorflow-0.9\"))' Datasets ImageClassificationModels\n",
         "print(\"\\u{001B}[2J\")"
       ],
-      "execution_count": 1,
-      "outputs": [
-        {
-          "output_type": "stream",
-          "text": [
-            "\r\n"
-          ],
-          "name": "stdout"
-        }
-      ]
+      "execution_count": 0,
+      "outputs": []
     },
     {
       "cell_type": "code",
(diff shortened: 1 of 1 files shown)


BradLarson:
Great job on the profiling. I had forgotten that we'd done the same for other internal models to avoid the use of `scalarized()` and its performance consequences. This also simplifies the training loop code.

Would it be possible to clean the execution state of your notebook, removing the output from the cells and their execution order? I'd like for the notebook to start blank, and it would be preferable for the diff here to only be the code that was changed.

mikowals:
I removed the outputs and execution state.  I did not realise those got saved as part of the notebook.

docs/site/tutorials/introducing_x10.ipynb
```diff
@@ -301,6 +302,50 @@
...
       "outputs": []
     },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "cwNDwzS2QgS1",
+        "colab_type": "text"
+      },
+      "source": [
+        "Now, we wiil implement basic progress tracking and reporting.  All data is kept as tensors on the same device where training is run and calls `scalarized()` only during reporting to avoid unneccesary materialization of lazy tensors."
```
BradLarson, line 312:
```suggestion
        "Now, we will implement basic progress tracking and reporting.  All intermediate statistics are kept as tensors on the same device where training is run and `scalarized()` is called only during reporting. This will be especially important later when using X10, because it avoids unnecessary materialization of lazy tensors."
```

docs/site/tutorials/introducing_x10.ipynb
```diff
@@ -301,6 +302,50 @@
...
+        "colab_type": "code",
+        "colab": {}
+      },
+      "source": [
+        "struct Statistics {\n",
+        "    var correctGuessCount = Tensor<Int32>(0, on: Device.default)\n",
+        "    var totalGuessCount = Tensor<Int32>(0, on: Device.default)\n",
+        "    var totalLoss = Tensor<Float>(0, on: Device.default)\n",
+        "    var batches: Int = 0\n",
+        "    var accuracy: Float { Float(correctGuessCount.scalarized()) / Float(totalGuessCount.scalarized()) * 100 } \n",
```
BradLarson, line 328: I'd recommend expanding this out onto the next two lines, to avoid the line length exceeding the width of the cell.

docs/site/tutorials/introducing_x10.ipynb
mikowals: Done.

docs/site/tutorials/introducing_x10.ipynb
mikowals: Done.

docs/site/tutorials/introducing_x10.ipynb
```diff
@@ -301,6 +302,52 @@
...
       "outputs": []
     },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "cwNDwzS2QgS1",
+        "colab_type": "text"
+      },
+      "source": [
+        "\"Now, we will implement basic progress tracking and reporting.  All intermediate statistics are kept as tensors on the same device where training is run and `scalarized()` is called only during reporting. This will be especially important later when using X10, because it avoids unnecessary materialization of lazy tensors.\""
```
BradLarson, line 312:
```suggestion
        "Now, we will implement basic progress tracking and reporting.  All intermediate statistics are kept as tensors on the same device where training is run and `scalarized()` is called only during reporting. This will be especially important later when using X10, because it avoids unnecessary materialization of lazy tensors."
```

mikowals:
Thanks for bearing with me as I add silly mistakes while hurrying commits.  I found your suggestions all very useful.

PR #502, opened 2020-07-17
Add notebooks about value semantics and AD

saeta:


saeta:
You can run these notebooks in the browser while this PR is open by going to:

 - https://colab.research.google.com/github/tensorflow/swift/blob/add-notebooks/notebooks/value_semantics/01_value_semantics_in_swift.ipynb
 - https://colab.research.google.com/github/tensorflow/swift/blob/add-notebooks/notebooks/value_semantics/02_value_semantics_and_autodiff.ipynb

Once this PR is merged, you'll need to update the paths (and I'll send a subsequent PR to add a `README.md` into this repository to make it easy for people to find the links).

notebooks/value_semantics/01_value_semantics_in_swift.ipynb
```diff
@@ -0,0 +1,949 @@
...
+    "    f.incrementByOne()  // Compile error\n",
+    "}\n",
+    "testFunc2()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Mutating functions are convenient in a variety of circumstances. For example, when using an iterator."
```
dabrahams, line 260: <strike>I have a small concern about using iterator to demonstrate value semantics, since copyability is a key element that makes value semantics interesting, just like mutability.  A noncopyable (always-uniquely-referenced) mutable reference type also has value semantics.  And copyability is ultimately the wrong model for iterator.  But maybe</strike> if they are thinking of the iterator as traversing a multi-pass thing, it's just fine

notebooks/value_semantics/01_value_semantics_in_swift.ipynb
```diff
@@ -0,0 +1,949 @@
...
+    "Said another way, these properties of Swift enable high performance execution without sacrificing value semantics.\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Early dropping of references\n",
+    "\n",
+    "Swift does not have a defined point of destruction for local variables, and thus the compiler is free to destroy variables early. Swift doesn't specify the memory layout of a local stack frame, which allows for a variety of optimizations, such as hoisting values to registers or tearing them apart. Below are a few code examples of this."
```
dabrahams, line 431:
```suggestion
    "Unlike C++ Swift does not mandate that local variables live until the end of their scope, so the compiler is free to destroy variables when they are no longer used.  That helps preserve the uniqueness property that allows efficient in-place mutation. Also, swift doesn't specify the memory layout of a local stack frame, which allows for a variety of optimizations, such as hoisting values to registers or tearing them apart. A few examples:"
```

notebooks/value_semantics/01_value_semantics_in_swift.ipynb
```diff
@@ -0,0 +1,949 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 18,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "/// An Int, but implemented using the same lazy copy-on-write techniques as\n",
+    "/// arrays (and Tensors) for demonstration purposes (makes copies explicit).\n",
+    "public struct CoWInt {\n",
```
dabrahams, line 442: IMO this example isn't illustrating anything about exploding values/local stack frames as claimed above, and keeping the copy count in the instance makes for a complex example that's hard for users to analyze.  We should discuss how to improve this part.

One idea: maybe it would be better simply to have `CoWInt` print a message when it makes a copy

notebooks/value_semantics/01_value_semantics_in_swift.ipynb
```diff
@@ -0,0 +1,949 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 34,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Simultaneous accesses to 0x7fc44a77c420, but modification requires exclusive access.\r\n",
```
dabrahams, line 881: This output shows a runtime detection, which is not what the example says it illustrates.

notebooks/value_semantics/01_value_semantics_in_swift.ipynb
```diff
@@ -0,0 +1,949 @@
...
+     "traceback": [
+      "Current stack trace:",
+      "\tframe #4: \u001b[33m0x00007fc44a772a7c\u001b[0m $__lldb_expr199`closure #1 in (aCopy=<unavailable>) at \u001b[36m<Cell 34>\u001b[0m:\u001b[33m3\u001b[0m:\u001b[33m7\u001b[0m",
+      "\tframe #5: \u001b[33m0x00007fc44a791fac\u001b[0m $__lldb_expr184`Int.increment(amount=0x0000000000000000, self=4) at \u001b[36m<Cell 31>\u001b[0m:\u001b[33m4\u001b[0m:\u001b[33m19\u001b[0m",
+      "\tframe #6: \u001b[33m0x00007fc44a772a09\u001b[0m $__lldb_expr199`main at \u001b[36m<Cell 34>\u001b[0m:\u001b[33m2\u001b[0m:\u001b[33m3\u001b[0m"
+     ]
+    }
+   ],
+   "source": [
+    "// Globals treated specially, so exclusivity detected at runtime.\n",
```
dabrahams, line 901:
```suggestion
    "// Mutable globals are shared mutable state, so exclusivity detected at runtime.\n",
```

notebooks/value_semantics/01_value_semantics_in_swift.ipynb
saeta: I think that would be a good improvement.

notebooks/value_semantics/01_value_semantics_in_swift.ipynb
saeta: In the JSON format, the output comes before the cell source. Are you using the ReviewNB extension?

dabrahams: The wha?

PR #507, opened 2020-07-26
[BitstreamParser] Fix parsing end block directive in block info block

owenv:
I was running some experiments with the BitstreamParser and noticed it wasn't aligning the stream when at the end of the BLOCKINFO block.


owenv:
@googlebot I signed it!

PR #515, opened 2020-08-14
Add documents describing the modern layer API requirements and prototypes

shadaj:


8bitmp3:
@shadaj Is it ok if I provide some suggestions?

shadaj:
@8bitmp3 sure, any feedback is definitely welcome!

PR #516, opened 2020-08-19
model training walkthrough minor improvements for clarity

garymm:
contentsOfCSV -> contentsOf. CSV is already in the function name so this
arg name seems needlessly verbose.

Use local variable names in comments about "This should never happen".
Saying "we construct <param name>" was a bit confusing to me, especially
since the param name in this case is also the name of the python module.

Print tensor shapes. I found this helpful to understand the process.


garymm:
@googlebot I signed it!

PR #525, opened 2020-08-26
fix import branch

garymm:


BradLarson:
Looking at this now, the main branch [did get updated](https://github.com/tensorflow/swift/commit/de939ff84b6ce64db2b1618c279a4f3e15013d93) to have this point to 0.11, so I think we're good in that regard. As a result, this PR looks to be empty right now. Sorry for the mistake in my previous reply.

PR #529, opened 2020-09-02
[model training walkthrough] update link to optimization algorithms

garymm:


docs/site/tutorials/model_training_walkthrough.ipynb
```diff
@@ -769,7 +769,7 @@
     "  </td></tr>\n",
     "</table>\n",
     "\n",
-    "Swift for TensorFlow has many [optimization algorithms](https://github.com/rxwei/DeepLearning/blob/master/Sources/DeepLearning/Optimizer.swift) available for training. This model uses the SGD optimizer that implements the *[stochastic gradient descent](https://developers.google.com/machine-learning/crash-course/glossary#gradient_descent)* (SGD) algorithm. The `learningRate` sets the step size to take for each iteration down the hill. This is a *hyperparameter* that you'll commonly adjust to achieve better results."
+    "Swift for TensorFlow has many [optimization algorithms](https://www.tensorflow.org/swift/api_docs/Classes) available for training. This model uses the SGD optimizer that implements the *[stochastic gradient descent](https://developers.google.com/machine-learning/crash-course/glossary#gradient_descent)* (SGD) algorithm. The `learningRate` sets the step size to take for each iteration down the hill. This is a *hyperparameter* that you'll commonly adjust to achieve better results."
```
BradLarson, line 772: The spiritual successor to the original link is here: https://github.com/tensorflow/swift-apis/tree/master/Sources/TensorFlow/Optimizers . Would we want to point to that, instead of documentation for all classes? At some point, we may even move optimizers away from being classes to being structs, at which point they'd disappear from this documentation list.

garymm: I personally think linking to documentation is better than linking to source code, but either way would be an improvement. Let me know what you prefer, or just make a suggested edit.

BradLarson, line 772:
```suggestion
    "Swift for TensorFlow has many [optimization algorithms](https://github.com/tensorflow/swift-apis/tree/master/Sources/TensorFlow/Optimizers) available for training. This model uses the SGD optimizer that implements the *[stochastic gradient descent](https://developers.google.com/machine-learning/crash-course/glossary#gradient_descent)* (SGD) algorithm. The `learningRate` sets the step size to take for each iteration down the hill. This is a *hyperparameter* that you'll commonly adjust to achieve better results."
```

Sorry for the slow reply, I think it's preferable to link to the optimizers themselves, due to the class documentation not being specific to optimizers by themselves. I've suggested that above. I agree that we should have more centralized documentation, but for now we can point here.

BradLarson (approved):
Thanks again for the fix!

PR #533, opened 2020-09-11
Format notebook tutorials with nbfmt

lamberta:
Format notebook tutorials with `nbfmt` from the [TensorFlow docs notebook tools](https://github.com/tensorflow/docs/tree/master/tools/tensorflow_docs/tools). This removes extra metadata for cleaner diffs (if used regularly). It also ensures that all cells have a `metadata.id` field (used by the new translation workflow).

```
python3 -m pip install -U --user git+https://github.com/tensorflow/docs

python3 -m tensorflow_docs.tools.nbfmt docs/site/
```

This PR keeps both 1 and 2 space indention levels for various notebooks, but normally I'd just default to 2 for everything (and example of editor differences that cause diff churn). I can set everything to 2, if desired.

Thanks!

### DIFF
--- a/docs/site/tutorials/a_swift_tour.ipynb
+++ b/docs/site/tutorials/a_swift_tour.ipynb
@@ -3,7 +3,6 @@
   {
    "cell_type": "markdown",
    "metadata": {
-    "colab_type": "text",
     "id": "QdsgF9Mxj0gr"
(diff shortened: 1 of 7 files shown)


dan-zheng (approved):
Thank you!

PR #555, opened 2020-11-06
Add 0.12 RCs

texasmichelle:


ematejska (approved):
Updated the version of the candidates and approved.

PR #571, opened 2020-11-18
Promote v0.12-rc2 to stable

texasmichelle:


ematejska (approved):
Looks good.  Do the Xcode requirements change for building?  I think I remember you installing a new version of Xcode and if so should that be updated in the installation page?

PR #594, opened 2021-01-05
Point Ubuntu nightly toolchains at the new stock versions

BradLarson:


dan-zheng (approved):
LGTM if the stock toolchains have been verified to work!

BradLarson:
@dan-zheng - They have, at least in all my tests and benchmarks on CPU and CUDA. This should give an opportunity for broader testing within the community, to see if we missed anything.

PR #598, opened 2021-01-20
Remove python3, python3-pip and python3-setuptools install. They are …

yashk2810:
…already available in the image


dan-zheng (approved):
Thank you!

yashk2810:
Thanks! Can you merge it so that I can run the job?

Issue #623, opened 2021-12-27
Resurrect S4TF


philipturner:
@compnerd some time after I finish reviving the Swift kernel for Google Colab ([link](https://github.com/philipturner/swift-colab)), I’m making a PR to this repo with major changes to the README. Since this would be going against Google's decision to archive it, I expect that you can't approve of the commit yourself. Is there any way we could contact people leading TensorFlow to get them to approve of un-archiving?

@BradLarson @rxwei @dan-zheng @lattner also pinging you, although @compnerd is the only person who seems to be able to make commits now.

philipturner:
For starters, the update date of the differentiable programming manifesto and implementation overview are out of date. Also, the guidance to try vanilla Swift in a broken Colab notebook is inaccurate.

philipturner:
We should wait to postpone merging the PR until after autodiff is stable enough to build swift-apis again. Also, I'll need to verify that GPU and TPU acceleration work on Google Colab. I don't see the need to wait for MetalXLA completion to do the un-archiving though.

Even though the PR has no chance of being merged right away, it's a good idea to work on a draft for a new README and on updating the Colab tutorials. It could happen on a branch or fork of S4TF, but I would like you to give feedback just as if it's merging into main.

philipturner:
On the Testing section of [Swift-Colab](https://github.com/philipturner/swift-colab)'s README, most tutorials that don't directly depend on TensorFlow are salvageable. It's just Custom Differentiation that's the problem (it was also discussed when transforming the [Differentiable Programming Implementation](https://github.com/apple/swift/blob/main/docs/DifferentiableProgrammingImplementation.md) into Markdown).

Since side-loading Swift on Google Colab requires a release toolchain, it requires importing my [Differentiation](https://github.com/philipturner/differentiation) package and replacing all instances of `_Differentiation` with `Differentiation`. Until autodiff is officially part of the Swift compiler, work on [swift-apis](https://github.com/tensorflow/swift-apis) may need to be done in a branch or fork.

PR #625, opened 2021-12-31
Update tutorials for new Colab support

philipturner:
With [Swift-Colab](https://github.com/philipturner/swift-colab), people can run the S4TF tutorials on Google Colab again. We need some minor tweaks to two, and major tweaks to the "Custom Differentiation" notebook that caused trouble while working on the [Differentiable Programming Implementation](https://github.com/apple/swift/blob/main/docs/DifferentiableProgrammingImplementation.md). No need to convert that one to DocC though!

This includes addressing the following changes regarding PythonKit and autodiff:
- `PythonLibrary.useVersion` causes a fatal error on Colab (possibly because [Swift-Jupyter](https://github.com/google/swift-jupyter) was translated from Python to Swift)
- Direct comparison operators on Python objects cause a compiler failure due to type ambiguity (solution: cast to `Bool`).
- `gradient(at:in:)` was renamed to `gradient(at:of:)`
- `withRecomputationInPullbacks(_:)` was silently removed (I couldn't find an explanation why after searching the commit history for an hour)
- `Differentiable.withDerivative` and `differentiableFunction(_:)` were removed.

For an analysis of specifically which notebooks are failing, look at the [very end](https://github.com/philipturner/swift-colab#testing) of the README on Swift-Colab.

---

The notebooks are not modified to include my side-loading workaround. You must manually copy the two code cells as instructed on Swift-Colab's README. I figure it is more appropriate to leave the side-loading part out of the notebooks.

The bug [SR-15675](https://bugs.swift.org/browse/SR-15675) was discovered and filed while making this PR (I assigned it to @BradLarson). In addition, this PR is related to the following comment: https://github.com/tensorflow/swift/issues/623#issuecomment-1003216770, but is not the "major changes" PR I am discussing there (it doesn't even modify the README).


philipturner:
@compnerd please review

